PR #13, opened 2021-09-21
Fix two syntax errors

fingolfin:
I wonder how this could slip through CI testing?


fingolfin:
Aha, so apparently the syntax errors are not detected because they do not actually affect the test results -- I assume that means the function where they are in is not tested.

But also I think we may need a "strict mode" for GAP that turns syntax warnings into errors, so that CI tests can catch them... I've submitted a GAP feature request for that now

ssiccha:
Ah, @maxhauck and @TristanPfersdorff, could you add tests for these code paths?

ssiccha (approved):

ssiccha:
Maybe that's superseded by #20?

fingolfin:
The issue is also fixed in #20 by @maxhauck, yes -- if that PR is merged quickly, this PR here can be closed. OTOH if that PR is likely to require more work, perhaps merge this PR and @maxhauck can rebase?

ssiccha:
We're rebasing #20.

PR #20, opened 2021-09-21
Unitary

### LINKED ISSUE #15: Add more unit tests
You already added some tests (good!) but there more functions that IMHO should have some basic tests to verify they return correct results. E.g. for
- [x] `SolveFrobeniusEquation`
- [x] `OddExtraspecialNormalizerInGL`
- [x] `SymplecticTypeNormalizerInGL`
- ... more to be added, see e.g. the coverage reports on https://app.codecov.io/gh/ssiccha/ClassicalMaximals/blob/main/gap/ReducibleMatrixGroups.gi and https://app.codecov.io/gh/ssiccha/ClassicalMaximals/blob/main/gap/ClassicalMaximals.gi

### LINKED ISSUE #24: Use `InvariantXForm`
When working with the forms left invariant by our classical groups, do not rely on the forms always being as documented in Section `50.2 Classical Groups` of the GAP reference manual. Instead use the there referenced functions `InvariantXForm` and our utility function `ChangeInvariantSesquilinearForm` (can't currently work with quadratic forms).

maxhauck:
All geometric subgroups of SU(d, q).

Includes the fixes in #13. Fixes #24. Fixes #15.


gap/ReducibleMatrixGroups.gi
```diff
@@ -251,37 +262,112 @@ function(d, q, k)
...
+            generator{[dHalf + 1]}{[1..kHalf]} :=
+                alpha ^ q * generatorOfSUSubspace{[kHalf + 1]}{[1..kHalf]};
+            generator{[dHalf]}{[d - kHalf + 1..d]} :=
+                generatorOfSUSubspace{[kHalf + 1]}{[kHalf + 2..k]};
+            generator{[dHalf + 1]}{[d - kHalf + 1..d]} :=
+                alpha ^ q * generatorOfSUSubspace{[kHalf + 1]}{[kHalf + 2..k]};
+            # finally, for the central 2x2-submatrix, we have to use all four
+            # relations above together
+            generator[dHalf][dHalf] := 
+                alpha * generatorOfSUSubspace[kHalf + 1][kHalf + 1]
```
fingolfin, line 287:
```suggestion
            generator[dHalf,dHalf] := 
                alpha * generatorOfSUSubspace[kHalf + 1,kHalf + 1]
```
... and similarly elsewhere

gap/ReducibleMatrixGroups.gi
```diff
@@ -251,37 +262,112 @@ function(d, q, k)
...
+            # now use e_{dHalf} = w_1 + beta ^ q * w_2
+            #         f_{dHalf} = alpha ^ q * w_1 - alpha * beta ^ q * w_2
+            generator{[dHalf]}{[1..kHalf]} :=
+                generatorOfSUSubspace{[kHalf + 1]}{[1..kHalf]}; 
+            generator{[dHalf + 1]}{[1..kHalf]} :=
+                alpha ^ q * generatorOfSUSubspace{[kHalf + 1]}{[1..kHalf]};
+            generator{[dHalf]}{[d - kHalf + 1..d]} :=
+                generatorOfSUSubspace{[kHalf + 1]}{[kHalf + 2..k]};
+            generator{[dHalf + 1]}{[d - kHalf + 1..d]} :=
+                alpha ^ q * generatorOfSUSubspace{[kHalf + 1]}{[kHalf + 2..k]};
```
fingolfin, line 283: The code in this function is really difficult to read and thus also difficult to verify that it is correct. I have several suggestions to improve on that.

First off, I think long "speaking" names are not a boon here, but rather a malus: if you rename `generator` to `g` and `generatorOfSUSubspace` to `h`, for example, this particular section can be rewritten as:
```suggestion
            g{[dHalf]}{[1..kHalf]}             := h{[kHalf + 1]}{[1..kHalf]}; 
            g{[dHalf + 1]}{[1..kHalf]}         := h{[kHalf + 1]}{[1..kHalf]} * alpha ^ q;
            g{[dHalf]}{[d - kHalf + 1..d]}     := h{[kHalf + 1]}{[kHalf + 2..k]};
            g{[dHalf + 1]}{[d - kHalf + 1..d]} := h{[kHalf + 1]}{[kHalf + 2..k]} * alpha ^ q;
```
which is *much* easier (for me, at least) to take in "at a glance".

(In addition, I would recommend avoiding extremely similar and at the same time long names like `generatorOfSUSubspace` and `generatorsOfSUSubspace`, as it is very easy to confuse them with each other -- the single letter difference is easily lost in the big letter sea).

gap/ReducibleMatrixGroups.gi
```diff
@@ -223,22 +223,33 @@ function(d, q, k)
...
         # (e_1, ..., e_{d / 2}, f_{d / 2}, ..., f_1 is the standard basis).
         #
+        # Note that, by construction of alpha and beta,
+        #       e_{dHalf} = w_1 + beta ^ q * w_2
+        #       f_{dHalf} = alpha ^ q * w_1 - alpha * beta ^ q * w_2
+        # Also by construction of alpha and beta, our standard unitary form
+        # applied to w_1 and w_1 or w_2 and w_2, respectively, each gives 1
+        # again, as needed.
+        #
         # The following matrices generate SU(k, q) x SU(d - k, q).
```
fingolfin, line 253: Reading the code diagonally, it seems that it takes a matrix; chops it into blocks; then uses the blocks (plus some additional sprinkles) to assemble a new matrix.

Perhaps using `BlockMatrix` would allow to make this much more readable?

gap/Utils.gi
```diff
@@ -146,7 +146,7 @@ function(type, alpha, q)
         for delta in GF(q) do
             polynomial := x ^ 2 + delta * x + alpha;
             if IsIrreducibleRingElement(R, polynomial) then
-                result := CoefficientsOfUnivariatePolynomial(Factors(S, polynomial)[1])[1];
+                result := -CoefficientsOfUnivariatePolynomial(Factors(S, polynomial)[1])[1];
```
fingolfin, line 149: ... and that's precisely the kind of issue which makes me suggest that small functions like `CoefficientsOfUnivariatePolynomial` should have their own unit tests :-)

gap/ReducibleMatrixGroups.gi
maxhauck: I did think of that, but according to the GAP Manual, `BlockMatrix` requires all blocks to have the same dimensions - which isn't the case here.

fingolfin: Ah indeed, it's been a long time I used it, I didn't realize it had this severe limitation. To bad. IMHO this kind of code looks a lot nicer in Julia, it has decent support for block matrices :-)

gap/ReducibleMatrixGroups.gi
fingolfin: @maxhauck ping?

tst/standard/ClassicalNormalizerMatrixGroups.tst
```diff
@@ -1,56 +1,84 @@
...
 true
 gap> n := 4;; q := 5;;
 gap> G := SymplecticNormalizerInSL(n, q);;
+gap> IsSubset(SL(n, q), GeneratorsOfGroup(G));
+true
 gap> Size(Group(GeneratorsOfGroup(G))) = Size(G);
 true
 gap> n := 6;; q := 4;;
 gap> G := SymplecticNormalizerInSL(n, q);;
+gap> IsSubset(SL(n, q), GeneratorsOfGroup(G));
```
fingolfin, line 15: This kind of testing doesn't scale too well. I would recommend to write a function `TestSymplecticNormalizerInSL := function(n,q) ... ` which performs all those tests and calls `error` with a meaningful message.

Then the rest of the `tst` file just becomes a bunch of nested loops calling this function with different values. And if you want to add a test in all cases, you only have to modify that one place.

Here is an example for this pattern in GAP: https://github.com/gap-system/gap/blob/master/tst/testinstall/grp/classic-forms.tst

gap/ReducibleMatrixGroups.gi
maxhauck: Well, I realised that there was no need to have the variable `generatorsOfSUSubspace` in the first place so it does not appear anymore in the current version of this code snippet, which neatly solves the "similar names"-issue.

Concerning the "speaking" names, I am a bit reluctant to change them ... - for one, I'm not quite so convinced that shorter names make the code significantly more readable here (the problem is rather the construction as such, which requires mixing and matching these matrix blocks) and secondly, I would prefer to leave them the way they are for consistency: We have used these long, "speaking" names everywhere so it would only make sense to stick to that here. 

Of course, if you feel very strongly about having those variable names changed, I can still do that :)

tst/standard/ClassicalNormalizerMatrixGroups.tst
maxhauck: When I first read this, I thought this was an excellent idea! 

However, thinking about this a bit, it might not be as straightforward as I thought: Many of the functions include casework with divisibility conditions and sometimes, there are just no "small" examples for such cases - so to test these somehow, I decided to just check if the generators are in the SU (or SL, respectively), i.e. I removed the `Size(Group(GeneratorsOfGroup(G))) = Size(G)` line since that simply took too long to compute. 

Bottom line: It would be nice to be able to manually control for which values both checks (generators in SU and right size) are run and for which values only one of them (i.e. generators in SU) is used. 

I guess one could still make this work by having several testing functions, such as `TestSymplecticNormalizerInSLSize` and `TestSymplecticNormalizerInSLDeterminant`, and calling each of them only on the particular values for which one wants that particular test to be run. - Would you recommend that?

ssiccha:
Parts of this PR were merged in via #22. The comments by @fingolfin will be adressed in this PR.

tst/standard/ClassicalNormalizerMatrixGroups.tst
fingolfin: I think the `Size` problem can and should be fixed by using `recog` to compute the size.

gap/Forms.gi
```diff
@@ -0,0 +1,90 @@
...
+end);
+
+ConjugateModule := function(M, q)
+  return GModuleByMats(List(MTX.Generators(M), A -> ApplyFunctionToEntries(A, x -> x ^ q)), 
+                       MTX.Field(M));
+end;
+
+# Assuming that the group G acts absolutely irreducibly, try to find a unitary
+# form which is G-invariant or prove that no such form exists.
+InstallGlobalFunction("UnitaryForm",
```
fingolfin, line 41: Have you considered using [`PreservedSesquilinearForms`](https://gap-packages.github.io/forms/doc/chap4.html#X84056A357E5447AF) from the GAP package `forms` instead?

gap/Forms.gi
```diff
@@ -0,0 +1,90 @@
...
+function(G)
+    local d, F, q, M, hermitianConjugateM, formMatrix, row, col, x, scalar;
+
+    d := DimensionOfMatrixGroup(G);
+    F := DefaultFieldOfMatrixGroup(G);
+    if not IsSquareInt(Size(F)) or not IsPrimePowerInt(Size(F)) then
```
fingolfin, line 47: The size of any finite field is a prime power, so this could be changed to
```suggestion
    if not IsSquareInt(Size(F)) then
```
However, it'd be slightly more efficient to just check that the degree over the prime field is even:
```suggestion
    if not IsFinite(F) or not IsEvenInt(DegreeOverPrimeField(F)) then
```

gap/SemilinearMatrixGroups.gi
```diff
@@ -102,3 +101,87 @@ function(n, q, s)
...
+    fi;
+
+    dimensionOfA := NumberRows(A);
+    result := NullMat(NumberRows(M) * dimensionOfA, 
+                      NumberColumns(M) * dimensionOfA, 
+                      DefaultFieldOfMatrix(A));
+
+    for i in [1..NumberRows(M)] do
+        for j in [1..NumberColumns(M)] do
+            if not M[i, j] = 0 * omega then
```
fingolfin, line 122:
```suggestion
            if not IsZero(M[i, j]) then
```

gap/Forms.gi
maxhauck: Yes and that works *reaaaaally* badly.

gap/Forms.gi
maxhauck: Done!

gap/Forms.gi
maxhauck: Maybe there is something wrong with `PreservedSesquilinearForms` as such? - Namely, I am guessing if it worked correctly, it should (despite using randomised methods) recognise a non-trivial preserved unitary form at least once in 10 000 trials ...?

gap/Forms.gi
```diff
@@ -0,0 +1,157 @@
...
+                                             field);
+    canonicalToNew := BaseChangeHomomorphism(BaseChangeToCanonical(newForm) ^ (-1), 
+                                             field);
+    return Group(canonicalToNew(gapToCanonical(GeneratorsOfGroup(group))));
+end);
+
+# Can only deal with sesquilinear forms, not with quadratic forms as of yet.
+# If <type> is one of "S", "O+", "O-" or "O" then <group> must have the
+# attribute InvariantBilinearForm.
+InstallGlobalFunction("ConjugateToGAPForm",
```
ssiccha, line 45: I suggest `ConjugateToStandardForm`. I think it makes sense to document this. Maybe this should go into the gap library? :thinking:

gap/Forms.gi
fingolfin: Sure, a bug is possible and this would ideally be reported via the `forms` issue tracker.

At the same time, please document this kind of reason in a code comment. Like, 
> not using FOO because it is slow/bugggy/INSERT-REASON as the example XYZ illustrates

gap/Forms.gi
fingolfin: Can't, is requires the forms package. But perhaps could be added there.

gap/Forms.gi
maxhauck: Comment added :) Where can I find the issue tracker? As you point out yourself in the GitHub repo of `forms`, the link on the `forms` homepage is broken ...

fingolfin: As usual it is part of the repo :-)

tst/standard/ClassicalNormalizerMatrixGroups.tst
ssiccha: This is handled in PR #36.

gap/Forms.gi
ssiccha: See #40.

gap/Forms.gi
ssiccha: See gap-packages/forms#22.

ssiccha (approved):

Issue #8, opened 2021-09-21
Use `SizeSL(d,q)`


fingolfin:
... instead of `Size(SL(d, q))`, and similar for `GL`, `PSL` -- this is more efficient.

Of course you need more, e.g. for `Sp` (which you sometimes write as `SymplecticGroup` ?), `PSp`, `SU`, ... -- I would recommend adding similar functions for these, two. In fact, this could, if you wanted to, turn into a pull request for GAP: there these sizes are all being computed in `grp/classic.gi`; one could add these functions in that file, and modify all the group constructors for these classical groups to call the corresponding `SizeFOO` functions... But that's an optional task :-)

PR #21, opened 2021-09-24
Size Functions

TristanPfersdorff:
Adds some Size functions to `gap/Utils.gi`, similar to `SizeSL(n, q)` as requested in #8 . These functions could probably use some tests in the future, also some more of them might need to be added in the future.


ssiccha (commented):
I haven't looked at the sizes in detail, I'll double-check these when I have time. Everything else looks good though.

ssiccha:
@maxhauck Could you double-check the sizes with the tables? If they are correct, approve the PR and I'll squash and merge.

maxhauck:
@ssiccha Everything looks fine as far as I can see except for one thing: The theorem in Bray, Holt, Roney-Dougal gives the formula for the size of GO(0, n, q) with n odd only under the condition that q is odd ... so the code has to include an exception for n odd and q even, which it currently does not. Namely, the case n odd and q even seems to be impossible in the orthogonal case (cf. for example Kleidman, Liebeck § 2.5 on p. 26) and so the function should return an error in that case; same thing of course for SO(0, n, q) with n odd, q even.

fingolfin:
> Namely, the case n odd and q even seems to be impossible in the orthogonal case

What do you mean with that? Impossible for what? Those groups of course exist (`GO(2n+1,2^e)` is isomorphic to `Sp(2n,2^e)`)

maxhauck:
> Those groups of course exist (`GO(2n+1,2^e)` is isomorphic to `Sp(2n,2^e)`)
My bad ... I just looked this up briefly then and Kleidman + Liebeck Prop. 2.5.1 seemed to suggest that orthogonal groups with q even require n even (and **even now**, knowing this is false, I still do not understand how else to interpret that Proposition, but anyway ...). 
The point remains, of course: There needs to be an exception for n odd, q even in which case `SizeGO` should simply call `SizeSp`. Also, I'm not sure if the index [GO : SO] is still 2 in the case q even ... I rather feel like we should have GO = SO for q even, so this has to be kept in mind for `SizeSO`.

gap/Utils.gi
```diff
@@ -168,3 +168,100 @@ function(m, n)
...
+        powerOfq := powerOfq * q;
+        sign := -sign;
+        result := result * (powerOfq + sign);
+    od;
+    return result;
+end);
+
+
+# Compute the size of GU(n, q) according to Table 1.3 in [1]
+InstallGlobalFunction("SizeGU",
```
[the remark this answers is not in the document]
ssiccha: I agree, that should be in here lest we forget about it.

fingolfin:
GO = SO if q is even is of course right, as then determinant -1 is equal to 1.

There seem to be some conflicts in this PR now, so it needs to be rebased

gap/Utils.gi
```diff
@@ -168,3 +168,100 @@ function(m, n)
...
     fi;
 end);
+
+
+# Compute the size of Sp(n, q) according to Theorem 1.6.22 in [1]
+InstallGlobalFunction("SizeSp",
+function(n, q)
+    local m, result, powerOfq, i;
+    if IsOddInt(n) then
+        ErrorNoReturn("Dimension <n> must be even but ", n, " was given.");
```
fingolfin, line 178: This line is not covered, and I actually would suggest to also test functions with invalid inputs to see how they are dealt with. However, this is an ultra premium extra special wish, so not important.

fingolfin (changes requested):
needs a rebase

gap/Utils.gi
```diff
@@ -180,6 +180,114 @@ function(m, n)
...
+        if IsEvenInt(n) then
+            ErrorNoReturn("for <epsilon> = ", epsilon, " the dimension <n> must be odd but ", n, " was given.");
+        fi;
+        if IsEvenInt(q) then
+            return SizeSp(n - 1, q);
+        fi;
+        return QuoInt(SizeGO(epsilon, n, q), 2);
+    else
+        return QuoInt(SizeGO(epsilon, n, q), Gcd(2, q - 1));
+    fi;
```
fingolfin, line 287: Can't you do this?
```suggestion
    return QuoInt(SizeGO(epsilon, n, q), Gcd(2, q - 1));
```
For `q` odd the `Gcd` is 2, so you get the same as before. And for `q` even, the Gcd is 1, and `SizeGO` already delegates to `SizeSp`.

fingolfin (approved):
Looks good to me. One minor remark.

ssiccha:
We're currently rebasing it.

ssiccha:
The conflicts we found until now are only due to updating the bibliography. They're easy to fix.

fingolfin (approved):

PR #22, opened 2021-09-29
Unitary by Maximilian

ssiccha:
- Unitary C1
- Unitary C2
- fix bug in C4SubgroupsSpecialLinearGroupGeneric
- write mat[i,j] instead of mat[i][j]
- fix ChangeSesquilinearForm
- Unitary C2 non-degenerate
- Unitary C2 in main function
- Unitary C4
- Unitary C4 main function
- faulty Unitary C6
- Unitary C6
- tests Unitary C6
- Unitary C6 in main function
- modify tests unitary
- modify tests & minor fixes
- Unitary C5, type GU and type Sp
- move GeneratorsOfOrthogonalGroup to Utils.gi
- Unitary C5 type GO


ssiccha:
These changes are a subset of #20 and were tested there. Thus I skipped tests here.

PR #32, opened 2021-10-01
Fix language

fingolfin:
"confer" means "grant, bestow" ;-)


ssiccha:
I have seen "confer" used as "refer" in some literature. I checked oed.com (see [this entry](https://www.oed.com/view/Entry/38737?redirectedFrom=confer#eid)), and that usage of "confer" is marked as obsolete.

ssiccha (approved):

PR #36, opened 2021-10-01
Some refactoring

### LINKED ISSUE #12: Be careful about relying on which precise generator sets GAP returns
I just saw this code in `C4SubgroupsSpecialLinearGroupGeneric`:
```gap
    generatorGLMinusSL := GL(n, q).1;
```
This encodes an assumption about the generators GAP will choose. But I don't think the GAP documentation makes any promises about them; so in future GAP versions, they could be different. Mind you, it is not likely right now (at least for `GL`), but possible. In general, I would at least add an `Assert` that somehow verifies this is the right (type of) element; or just explicitly write down whatever element you want. So, here, it could be this:
```gap
    generatorGLMinusSL := IdentityMat(n, GF(q));
    generatorGLMinusSL[1,1] := Z(q);
    generatorGLMinusSL := ImmutableMatrix( GF(q), generatorGLMinusSL );
```
Since you use it at a lot, maybe put it into a helper function.

maxhauck:
* Cache `GF(q)` etc.
* Fixes #12.

### DIFF
--- a/gap/ClassicalMaximals.gd
+++ b/gap/ClassicalMaximals.gd
@@ -19,3 +19,6 @@ DeclareGlobalFunction("ClassicalMaximalsGeneric");
 DeclareGlobalFunction("MaximalSubgroupClassRepsSpecialLinearGroup");
 
 DeclareGlobalFunction("MaximalSubgroupClassRepsSpecialUnitaryGroup");
+
+DeclareGlobalFunction("GLMinusSL");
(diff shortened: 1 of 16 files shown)


fingolfin:
rebase?

ssiccha:
> rebase?

Done. :)

ssiccha (approved):

PR #41, opened 2021-10-05
Start of almost simple in defining characteristic

danielrademacher:
Start of a file for the almost simple groups in defining characteristic (chapter 5 in the book by colva). In this PR we are also adding two functions to show how these groups can be generated in GAP.

### General


gap/AlmostSimpleDefiningCharacteristic.gi
```diff
@@ -0,0 +1,26 @@
+# Construction of group as in Table 5.6 row 4 from [BHR13]
+BindGlobal("l3qdim6",
+function(q)
+    local A, G, M, MM, S, T, general, w;
+    general := ValueOption("general");
```
ssiccha, line 5: The `ValueOption` abomination should be replaced by adding a variadic argument. This whole option business accesses and modifies a global state, from the docs `8 Options Stack`: "GAP  supports a global options system".

ssiccha:
I had the impression that, at least in the magma version, these functions share a lot of code. It may make sense to have one very abstract function, which then delegates to more specialized functions. E.g. I think many of the magma functions first build a tensor module "V tensor V", then compute the constituents, and then looks for a constituent with a given dimension. The only thing that changes between this part of these functions is the dimension we're looking for.

gap/AlmostSimpleDefiningCharacteristic.gi
fingolfin: Why does it even have to be variadic? Can't this function simply have two arguments, the second being `general`,l and people are free to pass in `false` for it?

BTW, while I agree options are problematic, they are often used to substitute for a feature GAP sadly is lacking: keyword arguments. In GAP you can end up with function calls like `func(true, true, false)` that can be rather confusing, compared to something like `func(general:=true, checks:=true, format_drive:=false)`

fingolfin: I guess what I tried to say with my last remark is: dang, we really should have keyword arguments in GAP ;-). I once started working on that but quickly decided to put it aside, because while I think implementing it would be fun, I really should be spending my time on other things *sigh*. But if anybody is ever interested in diving into the GAP kernel and implementing some language extensions, I'd be happy to teach you the ropes.

gap/AlmostSimpleDefiningCharacteristic.gi
```diff
@@ -0,0 +1,26 @@
+# Construction of group as in Table 5.6 row 4 from [BHR13]
+BindGlobal("l3qdim6",
```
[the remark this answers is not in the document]
ssiccha: Yepp, the current name is taken from magma. The abstract function could have a name like `ConstructDefiningCharacteristicEmbeddingOfAlmostSimpleGroup`, that's what's happening here. The function `l3qdim6` constructs one of these embeddings, namely the representation of an SL_3(q) on the symmetric-square module of dimension 6. There's a lot of these constructions which are extremely similar, which is why I'm suggesting something like `ConstructDefiningCharacteristicEmbeddingOfAlmostSimpleGroup` or maybe `ConstructDefiningCharacteristicEmbeddingOfAlmostSimpleGroupOnSymmetricPowerModule`

gap/AlmostSimpleDefiningCharacteristic.gi
```diff
@@ -0,0 +1,766 @@
...
+############################# OverSmallerField #####################################
+####################################################################################
+
+
+# The following functions are from the Recog Package.
+# Since some functions call other functions from the Recog Package, one
+# needs to load the Recog Package.
+
+
+# From the Recog Package but slightly modified (by DR).
```
ssiccha, line 19: Did you copy, paste that? Could we instead extend the function in recog such that you can also use it here?

fingolfin:
Any news on this? Would it help / make sense to meet to discuss this?

danielrademacher:
Anna and I haven't had time to continue working on the project until now. Hopefully we'll get back to it during the semester break.

PR #44, opened 2021-10-06
Fix #12

### LINKED ISSUE #12: Be careful about relying on which precise generator sets GAP returns

maxhauck:
_Description of your PR_

### General

### DIFF
--- a/gap/ClassicalNormalizerMatrixGroups.gi
+++ b/gap/ClassicalNormalizerMatrixGroups.gi
@@ -8,14 +8,13 @@ function(d, q)
 
     F := GF(q);
     zeta := PrimitiveElement(F);
-    A := Sp(d, q).1;
-    B := Sp(d, q).2;
+    AandB := ShallowCopy(GeneratorsOfGroup(Sp(d, q)));
     gcd := Gcd(d, q - 1);
     # generates the center of SL(d, q)
     C := zeta ^ QuoInt(q - 1, gcd) * IdentityMat(d, F);
 
     if IsEvenInt(q) or gcd / 2 = Gcd(q - 1, d / 2) then
-        result := Group([A, B, C]);
+        result := Group(Concatenation(AandB, [C]));
         # Size according to Table in [BHR13]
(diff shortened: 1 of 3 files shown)


gap/ClassicalNormalizerMatrixGroups.gi
```diff
@@ -8,14 +8,13 @@ function(d, q)
 
     F := GF(q);
     zeta := PrimitiveElement(F);
-    A := Sp(d, q).1;
-    B := Sp(d, q).2;
+    AandB := ShallowCopy(GeneratorsOfGroup(Sp(d, q)));
```
fingolfin, line 11: When reading this code later on, nobody will know about A and B, and thus wonder about this name.

I would recommend calling the variable `gens` instead (you could use `generators`, too, but in the GAP group theory code, the use of `gens` for a set of generators is ubiquitous). Then, later, replace `result := Group(Concatenation(AandB, [C]));` by
```gap
   Add(gens, C);
   result := Group(gens);
```
resp. replace `result := Group(Concatenation(AandB, [C, E]));` by
```gap
   Append(gens, [C, E]);
   result := Group(gens);
```

ssiccha (approved):
I'm OK with keeping the names as in the paper. We have done that in all constructor functions, if the paper provided names for each of the generators.

PR #48, opened 2021-10-06
Update references in ClassicalMaximals.gi as requested in #31

### LINKED ISSUE #31: Don't use [1], [2], [3] for references
These can very easily break, when someone edits the bibliography or imports your code into another package / into GAP itself, or whatever.

Instead, I would use something like this:
```
# [BHR13]  J. M. Bray, D. F. Holt, C. M. Roney-Dougal. "The Maximal Subgroups of the
#          Low-Dimensional Finite Classical Groups." Cambridge UP, 2013.
# [HR05]   D. F. Holt, C. M. Roney-Dougal. "Constructing Maximal Subgroups of
#          Classical Groups." LMS Journal of Computation and Mathematics, vol. 8,
#          2005, pp. 46-79.
# [KL90]   P. Kleidman, M. Liebeck. "The Subgroup Structure of the Finite
#          Classical Groups." Cambridge UP, 1990.
```
And then match it in your source code comments.

maxhauck:
Fixes #31.
 
### General

### DIFF
--- a/gap/ClassicalMaximals.gi
+++ b/gap/ClassicalMaximals.gi
@@ -1,14 +1,7 @@
 #
 # ClassicalMaximals: Maximal subgroups of classical groups
 #
-# Code along the lines of:
-# [BHR13]   J. M. Bray, D. F. Holt, C. M. Roney-Dougal. "The Maximal Subgroups of the
-#           Low-Dimensional Finite Classical Groups." Cambridge UP, 2013.
-# [HR05]       D. F. Holt, C. M. Roney-Dougal. "Constructing Maximal Subgroups of
-#           Classical Groups." LMS Journal of Computation and Mathematics, vol. 8,
-#           2005, pp. 46-79.
-# [KL90]       P. Kleidman, M. Liebeck. "The Subgroup Structure of the Finite
-#           Classical Groups." Cambridge UP, 1990.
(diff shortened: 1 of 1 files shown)


gap/ClassicalMaximals.gi
```diff
@@ -2,13 +2,13 @@
...
 #
 # Code along the lines of:
-# [1]   J. M. Bray, D. F. Holt, C. M. Roney-Dougal. "The Maximal Subgroups of the
-#       Low-Dimensional Finite Classical Groups." Cambridge UP, 2013.
-# [2]   D. F. Holt, C. M. Roney-Dougal. "Constructing Maximal Subgroups of
-#       Classical Groups." LMS Journal of Computation and Mathematics, vol. 8,
-#       2005, pp. 46-79.
-# [3]   P. Kleidman, M. Liebeck. "The Subgroup Structure of the Finite
-#       Classical Groups." Cambridge UP, 1990.
+# [BHR13]   J. M. Bray, D. F. Holt, C. M. Roney-Dougal. "The Maximal Subgroups of the
```
ssiccha, line 5: We don't need these anymore

maxhauck: Done

ssiccha (approved):

Issue #54, opened 2021-10-06
Revise MatrixGroup


fingolfin:
... according to the comments I left on the original PR #56

fingolfin:
I have no idea what I had in mind here, even after re-reading through all comments on PR #56 twice.

PR #55, opened 2021-10-06
Revise MatrixGroup and MatrixGroupWithSize

maxhauck:
... according to #54. 

### General


maxhauck:
Changes included in #56.

PR #56, opened 2021-10-06
Revise MatrixGroup and use it everywhere

maxhauck:
- Replace all uses of `Group` by `MatrixGroup` or `MatrixGroupWithSize`, respectively.
- Add tests for `MatrixGroup` and `MatrixGroupWithSize`

### General


gap/ClassicalNormalizerMatrixGroups.gi
```diff
@@ -78,7 +77,7 @@ end);
 BindGlobal("OrthogonalNormalizerInSL",
 function(epsilon, d, q)
     local F, generatingScalar, zeta, generatorsOfOrthogonalGroup, generators,
-    result, i1, DEpsilon, EEpsilon, X, W, i2, k;
+    result, i1, DEpsilon, EEpsilon, X, W, i2, k, size;
```
ssiccha, line 80: We don't need `result` anymore, right?

gap/ClassicalNormalizerMatrixGroups.gi
```diff
@@ -14,32 +14,32 @@ function(d, q)
     C := zeta ^ QuoInt(q - 1, gcd) * IdentityMat(d, F);
 
     if IsEvenInt(q) or gcd / 2 = Gcd(q - 1, d / 2) then
-        result := Group(Concatenation(AandB, [C]));
-        # Size according to Table in [BHR13]
-        SetSize(result, gcd * SizePSp(d, q));
+        # Size according to Table 2.11 in [BHR13]
+        size := gcd * SizePSp(d, q);
+        generators := Concatenation(AandB, [C]);
```
ssiccha, line 19: This can be moved to where `C` is defined.

gap/Utils.gd
```diff
@@ -27,7 +27,7 @@ DeclareGlobalFunction("GeneratorsOfOrthogonalGroup");
 #! @Description
 #! Return a matrix group <C>G</C> over the field <A>F</A> generated by the matrices from
 #! the list <A>gens</A>; the generators of <C>G</C> are immutable matrices over
-#! <A>F</A>. The attribute <Ref Func="DefaultFieldOfMatrixGroup"/> for the matrix
+#! <A>F</A>. The attribute <Ref Attr="DefaultFieldOfMatrixGroup"/> for the matrix
 #! group constructed will be <A>F</A>.
```
fingolfin, line 31: This is wrong, even if `F` is finite: it (currently) only holds for fields with <= 256 elements, as then `ImmutableMatrix` will convert the matrices into "compressed" matrices which store the field explicitly. Indeed:
```
gap> F:=GF(257^2);
GF(257^2)
gap> G:=MatrixGroup(F, [IdentityMat(2,F)]);
Group([ [ [ Z(257)^0, 0*Z(257) ], [ 0*Z(257), Z(257)^0 ] ] ])
gap> DefaultFieldOfMatrixGroup(G);
GF(257)
```

fingolfin: OK, I see you changed the code since I wrote this, so I guess the comment above is moot...

gap/ClassicalNormalizerMatrixGroups.gi
```diff
@@ -65,10 +65,9 @@ function(d, q)
         od;
     fi;
 
-    result := Group(gens);
     # Size according to Table 2.11 in [BHR13]
-    SetSize(result, SizeSU(d, q0) * Gcd(q0 - 1, d));
-    return result;
+    size := SizeSU(d, q0) * Gcd(q0 - 1, d);
+    return MatrixGroupWithSize(F, gens, size);
```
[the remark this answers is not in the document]
ssiccha: Hmm, I never thought about it that way. I actually like returning the result of this function more than "define result, then call `SetSize`, then return".

gap/Utils.gi
```diff
@@ -363,15 +363,32 @@ function(epsilon, n, q)
...
+                                  and HasGeneratorsOfMagmaWithInverses
+                                  and IsFinitelyGeneratedGroup
+                                  and HasIsEmpty
+                                  and IsFinite,
+                          gens, false, true);
+    gens := List(Immutable(gens), g -> ImmutableMatrix(F, g));
+
+    G := rec();
+    ObjectifyWithAttributes(G, typ, GeneratorsOfMagmaWithInverses, AsList(gens));
+    SetDefaultFieldOfMatrixGroup(G, F);
```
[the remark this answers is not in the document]
ssiccha: But I think that, until then this can be merged such that everyone can start using the different argument order of MatrixGroup.

ssiccha (approved):

ssiccha:
I can either merge this after lunch and we open a separate PR to make true on the promise for `DefaultFieldOfMatrixGroup`, or we wait until everything is resolved and merge then. I'd do the first option.

PR #68, opened 2021-10-07
Symplectic C1

TristanPfersdorff:
Add the necessary functions for the C1 subgroups in the symplectic case. The codecov bot will be happy once we add the symplectic main function

### General


gap/ReducibleMatrixGroups.gi
```diff
@@ -358,3 +358,104 @@ function(d, q, k)
...
+    if k >= d / 2 then
+        ErrorNoReturn("<k> must be less than <d> / 2 but <k> = ", k, 
+        " and <d> = ", d);
+    fi;
+
+    field := GF(q);
+    gens := [];
+    twok := 2 * k;
+
+    # These generators are bidagonal block matrices
```
fingolfin, line 439:
```suggestion
    # These generators are diagonal block matrices
```
Or perhaps you meant something else, but I am not sure what?

gap/ReducibleMatrixGroups.gi
```diff
@@ -358,3 +358,104 @@ function(d, q, k)
...
 end);
+
+
+# Construction as in Proposition 4.3 of [HR05]
+BindGlobal("SpStabilizerOfIsotropicSubspace",
+function(d, q, k)
+    local field, gens, I, J, GLgen, Xi, Spgen, Yi;
+
+    if IsOddInt(d) then
+        ErrorNoReturn("<d> must even but ", d, " was given.");
```
fingolfin, line 369: This line is not covered. Of course, it's an error, and you are careful to not trigger errors in your tests.

That said, in general I've come to find it useful to explicitly also add tests for functions with incorrect arguments, to verify that error handling is in place and working. As a bonus side effect, this also allows to get code coverage (closer) to 100%. It also helps to find things like formatting mistakes in error messages.

Note a requirement, thought, just a suggestion.

gap/ReducibleMatrixGroups.gi
TristanPfersdorff: It's supposed to be block matrices with blocks on both diagonals and zeros elsewhere, but I'm not sure if "bidiagonal" is common terminology, so I'll clarify the comment.

TristanPfersdorff:
I've implemented all suggested changes including clarifying some comments and rebasing. The errors (e.g. for odd dimension) are now also checked.

gap/ClassicalMaximals.gi
```diff
@@ -827,3 +827,17 @@ function(n, q, classes...)
 
     return maximalSubgroups;
 end);
+
+
+BindGlobal("C1SubgroupsSymplecticGroupGeneric",
+function(n, q)
+    local listOfks, result;
+    listOfks := [1..QuoInt(n, 2) - 1];
```
fingolfin, line 835: This is not covered, but my understanding this is because this function will only be called in a future PR, correct?

TristanPfersdorff: Correct. Theses PRs are only for the individual Classes, once they are all done we will call the generic functions from a function along the lines of `MaximalSubgroupClassRepsSymplecticGroup`. The generic functions will only be called from there.

fingolfin (approved):
Looks good to me, thanks!

PR #70, opened 2021-10-08
Add functions recognising invariant symplectic & symmetric bilinear forms

maxhauck:
- Add functions `BilinearForm`, `SymplecticForm` and `SymmetricBilinearForm`
- Add usage of stored invariant sesquilinear form to `UnitaryForm`
- Adjust `ConjugateToSesquilinearForm` to the aforementioned changes
- Add an exception to `ConjugateToSesquilinearForm` for the case of 1x1-matrices as the `Forms` package has a bug in that case

### General


gap/Forms.gi
```diff
@@ -1,56 +1,77 @@
...
-    fi;
-    gapToCanonical := BaseChangeHomomorphism(BaseChangeToCanonical(gapForm), 
-                                             field);
-    canonicalToNew := BaseChangeHomomorphism(BaseChangeToCanonical(newForm) ^ (-1), 
-                                             field);
-    result := MatrixGroup(field, canonicalToNew(gapToCanonical(GeneratorsOfGroup(group))));
-    if HasSize(group) then
-        SetSize(result, Size(group));
+    # The Form package has a bug for d = 1 so we need to make this exception
+    if not d = 1 then
```
fingolfin, line 36:
```suggestion
    if d <> 1 then
```

gap/Forms.gi
```diff
@@ -1,56 +1,77 @@
...
+        gapToCanonical := BaseChangeHomomorphism(BaseChangeToCanonical(gapForm), 
+                                                 field);
+        canonicalToNew := BaseChangeHomomorphism(BaseChangeToCanonical(newForm) ^ (-1), 
+                                                 field);
+        result := MatrixGroup(field, canonicalToNew(gapToCanonical(GeneratorsOfGroup(group))));
+    
+        # Set useful attributes
+        if HasSize(group) then
+            SetSize(result, Size(group));
+        fi;
```
fingolfin, line 53: You can transfer much more information at once
```suggestion
        UseIsomorphismRelation(group, result);
```

gap/Forms.gi
fingolfin: Also, how about handling the case were the forms are identical a bit more efficiently?
```suggestion
    if gapForm = newForm then
      # nothing to be done
      result := group;
    elif d <> 1 then
```

gap/Forms.gi
```diff
@@ -1,56 +1,77 @@
...
+    else
+        # replaces the Witt index check above
+        if (gramMatrix[1, 1] = 0 and formMatrix[1, 1] <> 0) 
+           or (gramMatrix[1, 1] <> 0 and formMatrix[1, 1] = 0) then
```
fingolfin, line 57: Are these matrices really using integer entries? That's weird... But if not, then comparing to `0` will always yield that they are unequal.

You can also simplify the code further:
```suggestion
        if IsZero(gramMatrix[1, 1]) <> IsZero(formMatrix[1, 1]) then
```
or even
```suggestion
        if IsZero(gramMatrix) <> IsZero(formMatrix) then
```

gap/Forms.gi
```diff
@@ -1,56 +1,77 @@
...
+            ErrorNoReturn("The form preserved by <group> must be similar to the",
+                          " form described by the Gram matrix <gramMatrix>.");
+        fi;
+        result := group;
```
fingolfin, line 61: Codecov says the case d=1 is not covered by tests, so perhaps time to add one?

gap/Forms.gi
```diff
@@ -1,56 +1,77 @@
...
+    fi;
+
+    if type = "S" or type = "O" then
+        SetInvariantBilinearForm(result, rec(matrix := gramMatrix));
+    else
+        SetInvariantSesquilinearForm(result, rec(matrix := gramMatrix));
     fi;
```
fingolfin, line 68: So I guess this means even characteristic is not (yet) fully supported here, right? Because no quadratic form is set

gap/Forms.gi
```diff
@@ -198,12 +227,97 @@ function(G)
...
+    return fail;
+end);
+
+# Assuming that the group G acts absolutely irreducibly, try to find a
+#   * symplectic form (if <type> = S) or a 
+#   * symmetric bilinear form (if <type> = O)
+# which is G-invariant or prove that no such form exists.
+#
+# In general, this function should only be used if one can be sure that <G>
+# preserves a symplectic form (but one does not know which one).
```
fingolfin, line 246: Maybe also briefly mention why you are not using `PreservedForms` and friends.

gap/Forms.gi
maxhauck: Yes, this function only does bilinear forms, not quadratic forms, i.e. the orthogonal case only in odd characteristic

maxhauck:
It would be great to have this merged soon since it is needed for other functions ...

fingolfin (approved):

PR #74, opened 2021-10-30
Symplectic C6

maxhauck:
Add functions constructing C6 subgroups of Sp(n, q) and a corresponding generic function in the `ClassicalMaximals.gi`

Note: Lines not covered by the tests in `ExtraspecialNormalizerMatrixGroups.gi` are exclusively due to problems with RECOG, but I have checked these manually; the changes in `ClassicalMaximals.gi` will be tested once the symplectic main function is complete and we implement its tests.

### General


fingolfin:
Is the error in recog you are alluding to the same as https://github.com/gap-packages/recog/issues/295 ?

gap/ClassicalMaximals.gi
```diff
@@ -827,3 +827,58 @@ function(n, q, classes...)
...
+    factorisationOfn := PrimePowersInt(n);
+    r := factorisationOfn[1];
+    m := factorisationOfn[2];
+    generatorNormSpMinusSp := NormSpMinusSp(n, q);
+
+    # Cf. Table 4.6.B and the corresponding definition in [KL90]
+    if r = 2 and e = 1 then
+        extraspecialNormalizerSubgroup := ExtraspecialNormalizerInSp(m, q);
+        # Cf. Tables 3.5.C and 3.5.G in [KL90]
+        if (q - 1) = 0 mod 8 or (q - 7) = 0 mod 8 then
```
fingolfin, line 872: Could also write this (matter of taste, feel free to leave as-is)
```suggestion
        if q mod 8 in [1, 7] then
```

gap/ExtraspecialNormalizerMatrixGroups.gi
```diff
@@ -500,61 +536,108 @@ function(m, q, type...)
...
-                                    [U1, V1]);
+                                    [U1],
+                                    listOfVi,
+                                    listOfWi);
 
     else
-        # According to the Magma code, there is no need to take another
-        # generator instead of U1 if we cannot rescale it to determinant 1 - we
-        # simply discard U1 as a generator.
+        # Comparing with the Magma code, we can simply take
```
fingolfin, line 609: What do you mean here? Are you saying: "The Magma code does this, so we use this as justification for doing the same"? (that doesn't seem like a great argument, by the way ;-).

Or are you saying: "compared to the Magma code, we are better/different/whatever, and hence it suffices to do XYZ?

fingolfin (approved):

gap/ExtraspecialNormalizerMatrixGroups.gi
maxhauck: The former :D ... well, sadly the relevant paper does not really say much (or rather anything) concerning this particular case so I trusted the Magma code on this one - however, doing this did pass some sanity checks: (1) the group sizes do come out the right way and (2) this case is exactly where an exception occurs in the structure of these C6 subgroups according to the tables in [BHR13] (namely, they are a bit smaller), so it makes sense that this should suffice.

One could of course check "theoretically" that these generators are enough, but I am not sure if this maybe transcends our current mathematical horizon a bit ;)

maxhauck:
I guess I can merge this, @fingolfin?

fingolfin:
> - the tests just produced a ton of RECOG errors, which we currently handle by simply removing the error-causing tests until there are no errors anymore

That's bad! Please at least turn those into bug reports for recog. Else they won't (can't!!) be fixed

fingolfin:
There is a conflict now, which looks easy to resolve (basically two PRs were appending to the end of `gap/ClassicalMaximals.gi`). But I wonder if perhaps it would make sense to move the various cases C1, C3, etc. into separate source files, and only leave truly common stuff in `gap/ClassicalMaximals.gi`?

fingolfin:
I've manually resolved the conflict (via the web interface -- nice feature!) Assuming tests pass, we can merge it next.

PR #75, opened 2021-10-31
Symplectic C5

TristanPfersdorff:
Add the necessary functions for the C5 subgroups in the symplectic case. The codecov bot will be happy once we add the symplectic main function.

Edit: It appears that the checks have some really fishy issue due to RECOG, but I was only able to reproduce said issue once in a blue moon locally. Everything seems to work fine 99% of the time, so this must be some randomized-algorithm-stuff from RECOG.

### General


gap/ClassicalMaximals.gi
```diff
@@ -827,3 +827,17 @@ function(n, q, classes...)
...
+function(n, q)
+    local factorisation, p, e, degreeOfExtension, result;
+    
+    factorisation := PrimePowersInt(q);
+    p := factorisation[1];
+    e := factorisation[2];
+
+    # For each prime divisor of e, there is exactly one of these subgroups,
+    # so this is sufficient.
+    return List(PrimeDivisors(e), b -> SubfieldSp(n, p, e, QuoInt(e ,b)));
```
maxhauck, line 842: You have to remember conjugating the subgroup you constructed by an element of GSp \ Sp in order to get all the conjugacy classes in Sp, see for example ll. 833-834 at https://github.com/hulpke/maxtrans/blob/master/magma/classmax.m

gap/SubfieldMatrixGroups.gi
```diff
@@ -278,3 +278,45 @@ function(epsilon, d, q)
...
+
+    F := GF(p ^ e);
+    q0 := p ^ f;
+    b := QuoInt(e, f);
+    gens := List(GeneratorsOfGroup(Sp(d, q0)));
+
+    if e mod f <> 0 or not IsPrime(b) then
+        ErrorNoReturn("<f> must be a divisor of <e> and their quotient must be a prime but <e> = ", 
+                      e, " and <f> = ", f);
+    fi;
```
fingolfin, line 300: In general I'd try move input validation to the very start, before any work is done. Also, this checks for two conditions, and so I'd suggest to have two errors; that makes it slightly easier for a human to understand what's going on. I.e., together with what I wrote above:
```suggestion
    if e mod f <> 0 then
        ErrorNoReturn("<f> must be a divisor of <e>");
    fi;
    if not IsPrime(b) then
        ErrorNoReturn("The quotient of <f> by <e> must be a prime");
    fi;
```

gap/SubfieldMatrixGroups.gi
```diff
@@ -278,3 +278,45 @@ function(epsilon, d, q)
...
+    l := QuoInt(d, 2);
+    zeta := PrimitiveElement(F);
+    omega := PrimitiveElement(GF(q0));
+    zetaPower := zeta ^ - QuoInt(q0 + 1, 2);
+
+    # This matrix C preserves the form and is constructed to
+    # have determinant 1, but it is not in Sp(d, q0). Therefore it is
+    # our missing generator to extend Sp(d, q0) to a C5-subgroup, since
+    # C is in the Normalizer of Sp(d, q) of Sp(d, q0).
+    C := DiagonalMat(Concatenation(List([1..l], i -> omega * zetaPower), List([1..l], i -> zetaPower)));
```
fingolfin, line 317:
```suggestion
    C := DiagonalMat(Concatenation(ListWithIdenticalEntries(l, omega * zetaPower), ListWithIdenticalEntries(l, zetaPower)));
```

fingolfin (commented):
Some suggestions. Overall looks fine to me, modulo the remark by @maxhauck

TristanPfersdorff:
The proposed changes seem good to me, I have implemented them and rebased. As for the remark by @maxhauck, it seems like it suffices to call `ConjugatesInGeneralGroup` with his function `NormSpMinusSp` from the C6-PR (#74) in the right spot once that PR has been merged (maybe one could do some fancy git-stuff with this PR depending on #74, but I'm no expert).
I still don't really know what to make of that weird RECOG-error (`Error, Matrix INV: <mat> must be square (not 2 by 4)`); It seems to happen somewhat frequently when the test run for the first time, but it rarely occurs when re-running the tests (without reloading the package). I've encountered it a couple times now but I can't seem to make sense of it, since every matrix we use is square.

fingolfin (approved):

fingolfin (changes requested):
As with PR #74 this now has a conflict in `gap/ClassicalMaximals.gi` so this PR here can't be merged; also once PR #74 is merged we may have a fresh conflict. So yeah, perhaps move the C1/C2/... cases into separate files?

Also, tests are failing, that needs to be resolved first (I overlooked that).

tst/standard/SubfieldMatrixGroups.tst
```diff
@@ -71,3 +71,25 @@ gap> testsOrthogonalSubfieldSU := [[0, 3, 5], [0, 5, 3], [-1, 2, 3], [-1, 4, 3]]
...
+>   hasSize := HasSize(G);
+>   RECOG.TestGroup(G, false, Size(G));
+>   return IsSubset(Sp(n, p ^ e), GeneratorsOfGroup(G))
+>          and DefaultFieldOfMatrixGroup(G) = GF(p ^ e)
+>          and hasSize;
+> end;;
+gap> testsSubfieldSp := [[6, 2, 2, 1], [4, 3, 2, 1], [4, 3, 4, 2], [4, 7, 2, 1]];;
+gap> ForAll(testsSubfieldSp, TestSubfieldSp);
+true
+gap> SubfieldSp(3, 2, 2, 1);
```
fingolfin, line 90: I would suggest to add a comment that explains error handling is being tested (as opposed to a broken test file)
```suggestion

# test error handling
gap> SubfieldSp(3, 2, 2, 1);
```

PR #76, opened 2021-11-06
Symplectic C3

maxhauck:
The same note as in #74 applies: The changes in ClassicalMaximals.gi will be tested once the symplectic main function is complete and we implement its tests.

### General


gap/ClassicalMaximals.gi
```diff
@@ -827,3 +827,25 @@ function(n, q, classes...)
...
+
+    # symplectic type subgroups
+    for s in primeDivisorsOfn do
+        if IsEvenInt(n / s) then
+            Add(result, SymplecticSemilinearSp(n, q, s));
+        fi;
+    od;
+
+    # unitary type subgroups
+    if 2 in primeDivisorsOfn then
```
fingolfin, line 846:
```suggestion
    if IsEvenInt(n) then
```

gap/ClassicalMaximals.gi
```diff
@@ -827,3 +827,25 @@ function(n, q, classes...)
...
+
+    primeDivisorsOfn := PrimeDivisors(n);
+    result := [];
+
+    # symplectic type subgroups
+    for s in primeDivisorsOfn do
+        if IsEvenInt(n / s) then
+            Add(result, SymplecticSemilinearSp(n, q, s));
+        fi;
+    od;
```
fingolfin, line 843: I think this could be improved somewhat. Let me rephrase what you do: you are looping over all even divisors `d` (= `n/s`) of `n` and then do something with `n/d` (=`n/(n/s)`=`s`). But the even divisors of `n` either don't exist (if `n` is odd) or they are equal to the divisors of `n/2`, each multiplied by 2.

gap/ClassicalMaximals.gi
fingolfin: All in all this function seems to return the empty list when `n` is odd.

gap/SemilinearMatrixGroups.gi
```diff
@@ -194,3 +194,85 @@ function(d, q, s)
...
     return ConjugateToStandardForm(MatrixGroupWithSize(F, generators, size), "U");
 end);
+
+# Construction as in Proposition 6.4 of [HR05]
+BindGlobal("SymplecticSemilinearSp",
+function(d, q, s)
+    local F, gammaL1, As, Bs, m, omega, AandB, C, i, range, generators, size;
+    if d mod s <> 0 or not IsPrime(s) then
+        ErrorNoReturn("<s> must be prime and a divisor of <d> but <s> = ", s,
+                      " and <d> = ", d);
```
fingolfin, line 204:
```suggestion
        ErrorNoReturn("<s> must be prime and a divisor of <d>");
```
I don't think printing `s` and `d` is necessary here, as the user can just query them via the break loop once they hit that error, no?

gap/SemilinearMatrixGroups.gi
```diff
@@ -194,3 +194,85 @@ function(d, q, s)
...
+    fi;
+    if not IsEvenInt(QuoInt(d, s)) then
+        ErrorNoReturn("The quotient <d> / <s> must be even but <d> = ", d, 
+                      " and <s> = ", s);
+    fi;
+    F := GF(q);
+    gammaL1 := MatricesInducingGaloisGroupOfGFQToSOverGFQ(s, q);
+    # Let w be a primitive element of GF(q ^ s) over GF(q). Since As is the
+    # companion matrix of the minimal polynomial of w over GF(q), its
```
fingolfin, line 213: At this point, `As` is not even defined. Perhaps the line setting `As` should be moved before this comment? Or perhaps say something like
```suggestion
    # Let w be a primitive element of GF(q ^ s) over GF(q). The matrix As we define
    #  next is the companion matrix of the minimal polynomial of w over GF(q), hence its
```

gap/SemilinearMatrixGroups.gi
```diff
@@ -194,3 +194,85 @@ function(d, q, s)
...
+    if d mod 2 <> 0 then
+        ErrorNoReturn("<d> must be divisible by 2 but <d> = ", d);
+    fi;
+    if q mod 2 = 0 then
+        ErrorNoReturn("<q> must be odd but <q> = ", q);
+    fi;
+
+    F := GF(q);
+    gammaL1 := MatricesInducingGaloisGroupOfGFQToSOverGFQ(2, q);
+    # Let w be a primitive element of GF(q ^ 2) over GF(q). Since A2 is the
```
fingolfin, line 251: see above

gap/ClassicalMaximals.gi
fingolfin: Actually, here of course you only look at the *prime* divisors, so basically all you need is to let s iterate over the *odd* prime divisors. So unless I am missing something, one could do:
```suggestion
    if not IsEvenInt(n) then return []; fi;

    primeDivisorsOfn := PrimeDivisors(n);
    RemoveSet(primeDivisorsOfn, 2)
    result := [];

    # symplectic type subgroups
    for s in primeDivisorsOfn do
        Add(result, SymplecticSemilinearSp(n, q, s));
    od;
```
However, my original remark still applies to the C2 case in PR #77.

maxhauck: Be careful here: If n is divisible by 4, then s = 2 is still admissible (as n / s = n / 2 is still even) - so you can't just throw out s = 2 entirely. 

Of course, you could add an if-statement saying "throw out s = 2 if n is not divisible by 4, otherwise leave it in", but I think then the new syntax would not only be not much easier or clearer than the original one, but it would also obfuscate the intention behind it all (namely, that we just need to ensure that n / s is even).

gap/SemilinearMatrixGroups.gi
maxhauck: Well, so far, we always printed the relevant arguments in error messages - so you're suggesting to alter this paradigm?

gap/SemilinearMatrixGroups.gi
maxhauck: We have written comments explaining newly defined objects above the relevant object all over the code - e.g., there are multiple more instances of this in the `SemilinearMatrixGroups.gi`. Personally, I find it more logical to have a comment above the relevant code instead of below ... - of course, if you feel strongly about this, we can also change it.

gap/SemilinearMatrixGroups.gi
fingolfin: Yes, I think I also suggested this in other PRs before. i can also turn this into an issue if you prefer. The way I suggest is the usual way to do this in GAP

gap/SemilinearMatrixGroups.gi
fingolfin: Having the comment above is fine by me but then it must be clear that it references things below it, because we read code from top to bottom, not the other way around. For a 1 line comment this is always fine, for large comments less so. And here it mixes a reference to an object not yet defined with claims about that object not immediately clear from the surrounding code (one has to look at a function call *before* the comment and look at that function resp. its docs/comments

PR #80, opened 2021-11-17
Add START_TEST/STOP_TEST to all .tst files

fingolfin:
This resets the RNG state and thus makes test results easier to reproduce


fingolfin:
This actually "fixes" the tests, but simply because the RNG states are now fixed per .tst file.

PR #81, opened 2021-11-18
Don't use ForAll to run many tests in a row

fingolfin:
Also add some comments for disabled tests that explain why they are disabled.
Ideally all of these would include links to issues reporting the problem. I've only done this in a few cases, the others still need to be analyzed and possible issues be filed.

With this setup, it is much easier to pinpoint which tests precisely is failing; one also gets better (more fine granular) feedback on the progress of the tests.

### DIFF
--- a/tst/standard/ImprimitiveMatrixGroups.tst
+++ b/tst/standard/ImprimitiveMatrixGroups.tst
@@ -9,17 +9,17 @@ gap> TestImprimitivesMeetSL := function(args)
 >   G := ImprimitivesMeetSL(n, q, t);
 >   hasSize := HasSize(G);
 >   RECOG.TestGroup(G, false, Size(G));
->   return IsSubset(SL(n, q), GeneratorsOfGroup(G))
->          and DefaultFieldOfMatrixGroup(G) = GF(q)
->          and hasSize;
+>   Assert(0, IsSubset(SL(n, q), GeneratorsOfGroup(G)));
+>   Assert(0, DefaultFieldOfMatrixGroup(G) = GF(q));
+>   Assert(0, hasSize);
 > end;;
+gap> TestImprimitivesMeetSL([2, 3, 2]);
 #@if IsBound(CLASSICAL_MAXIMALS_RUN_BROKEN_TESTS)
-gap> testsImprimitivesMeetSL := [[2, 3, 2], [4, 8, 2], [6, 5, 3]];;
-#@else
-gap> testsImprimitivesMeetSL := [[2, 3, 2], [6, 5, 3]];;
+gap> TestImprimitivesMeetSL([4, 8, 2]); # FIXME: `Error, !!!`, see https://github.com/gap-packages/recog/issues/12
 #@fi
-gap> ForAll(testsImprimitivesMeetSL, TestImprimitivesMeetSL);
-true
+gap> TestImprimitivesMeetSL([6, 5, 3]);
+
+#
 gap> TestSUIsotropicImprimitives := function(args)
 >   local n, q, G, hasSize;
 >   n := args[1];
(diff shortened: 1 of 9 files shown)


tst/standard/ImprimitiveMatrixGroups.tst
```diff
@@ -9,35 +9,33 @@ gap> TestImprimitivesMeetSL := function(args)
...
+>   Assert(0, IsSubset(SL(n, q), GeneratorsOfGroup(G)));
+>   Assert(0, DefaultFieldOfMatrixGroup(G) = GF(q));
+>   Assert(0, hasSize);
 > end;;
+gap> TestImprimitivesMeetSL([2, 3, 2]);
 #@if IsBound(CLASSICAL_MAXIMALS_RUN_BROKEN_TESTS)
-gap> testsImprimitivesMeetSL := [[2, 3, 2], [4, 8, 2], [6, 5, 3]];;
-#@else
-gap> testsImprimitivesMeetSL := [[2, 3, 2], [6, 5, 3]];;
+gap> TestImprimitivesMeetSL([4, 8, 2]); # FIXME: `Error, !!!`, see https://github.com/gap-packages/recog/issues/12
```
fingolfin, line 18: For reference / linking issues: https://github.com/gap-packages/recog/issues/12

tst/standard/TensorProductMatrixGroups.tst
```diff
@@ -24,17 +30,18 @@ gap> TestTensorProductStabilizerInSU := function(args)
...
+>   Assert(0, hasSize);
 > end;;
+gap> TestTensorProductStabilizerInSU([2, 3, 2]);
+gap> TestTensorProductStabilizerInSU([2, 3, 3]);
+gap> TestTensorProductStabilizerInSU([2, 3, 4]);
 #@if IsBound(CLASSICAL_MAXIMALS_RUN_BROKEN_TESTS)
-gap> testsTensorProductStabilizerInSU := [[2, 3, 2], [2, 3, 3], [2, 3, 4], [2, 4, 3]];;
-#@else
-gap> testsTensorProductStabilizerInSU := [[2, 3, 2], [2, 3, 3]];;
+gap> TestTensorProductStabilizerInSU([2, 4, 3]); # FIXME: see https://github.com/gap-packages/recog/issues/302
```
fingolfin, line 41: For reference / linking issues: https://github.com/gap-packages/recog/issues/302

tst/standard/ReducibleMatrixGroups.tst
```diff
@@ -10,17 +10,17 @@ gap> TestSLStabilizerOfSubspace := function(args)
...
+>   Assert(0, DefaultFieldOfMatrixGroup(G) = GF(q));
+>   Assert(0, hasSize);
 > end;;
-#@if IsBound(CLASSICAL_MAXIMALS_RUN_BROKEN_TESTS) 
-gap> testsSLStabilizerOfSubspace := [[4, 3, 2], [3, 8, 2], [2, 7, 1]];;
-#@else
-gap> testsSLStabilizerOfSubspace := [[4, 3, 2], [2, 7, 1]];;
+gap> TestSLStabilizerOfSubspace([4, 3, 2]);
+#@if IsBound(CLASSICAL_MAXIMALS_RUN_BROKEN_TESTS)
+gap> TestSLStabilizerOfSubspace([3, 8, 2]); # FIXME: `Error, !!!`, see https://github.com/gap-packages/recog/issues/12
```
fingolfin, line 19: For reference / linking issues: https://github.com/gap-packages/recog/issues/12

maxhauck (approved):
This can be merged.

tst/standard/SubfieldMatrixGroups.tst
```diff
@@ -30,50 +30,47 @@ gap> TestUnitarySubfieldSU := function(args)
 >   G := UnitarySubfieldSU(n, p, e, f);
 >   hasSize := HasSize(G);
 >   RECOG.TestGroup(G, false, Size(G));
->   return IsSubset(SU(n, p ^ e), GeneratorsOfGroup(G))
->          and DefaultFieldOfMatrixGroup(G) = GF(p ^ (2 * e))
->          and hasSize;
+>   Assert(0, IsSubset(SU(n, p ^ e), GeneratorsOfGroup(G)));
+>   #Assert(0, DefaultFieldOfMatrixGroup(G) = GF(p ^ (2 * e))); # FIXME
```
fingolfin, line 34: @maxhauck can yo tell me what's going on here? Should it perhaps be like this?
```suggestion
>   Assert(0, DefaultFieldOfMatrixGroup(G) = GF(p ^ (2 * f)));
```
That at least works in the first two test cases. It fails the third one, though
```
gap> TestUnitarySubfieldSU([3, 5, 3, 1]);
Error, Assertion failure in
  DefaultFieldOfMatrixGroup( G ) = GF( p ^ (2 * f) ) at *stdin*:40 called from
<function "TestUnitarySubfieldSU">( <arguments> )
 called from read-eval loop at *stdin*:45
you may 'return;'
brk> DefaultFieldOfMatrixGroup( G );
GF(5^6)
brk> e;
3
brk> f;
1
```

maxhauck: No, I am pretty certain it _should be_ `GF(p ^ (2 * e))` indeed - however, this is again a problem with the base field of the matrix group: The group can be written over `GF(p ^ (2 * f))` and so GAP returns `GF(p ^ (2 * f))` in the first two cases instead of the much larger `GF(p^ (2 * e))` (which it should actually be).

fingolfin: OK, but then the tests perhaps should just be whether the `DefaultFieldOfMatrixGroup` is a subfield -- and that's already covered by the check in the line above. So I think it could just be removed.

PR #85, opened 2021-11-26
Implement Orthogonal C1 symplectic type

maxhauck:
Add the following functions:
- [x] `OmegaStabilizerOfNonSingularVector` in `ReducibleMatrixGroups.gi`
- [x] `SolveQuadraticEquation`, `FindGamma` and `FancySpinorNorm` in `Utils.gi`
- [x] `QuadraticForm` in `Forms.gi`

Modify the following functions:
- [x] `ReflectionMatrix` in `Utils.gi`
- [x] `ConjugateToSesquilinearForm` and `ConjugateToStandardForm` in `Forms.gi`

Tests have now been added as well, lines not covered are almost exclusively error messages.

### General

### DIFF
--- a/doc/ClassicalMaximals.bib
+++ b/doc/ClassicalMaximals.bib
@@ -31,10 +31,6 @@ MRREVIEWER = {A. S. Kondrat\cprime ev},
        URL = {https://doi.org/10.1112/S1461157000000899},
 }
 
-%%%%%%%%%%%%%%%%%%%%%%%%%%%
-% TODO
-% Add MRNUMBER, MRREVIEWER
-%%%%%%%%%%%%%%%%%%%%%%%%%%%
 @article {HR10,
     AUTHOR = {Holt, Derek F. and Roney-Dougal, Colva M.},
      TITLE = {Constructing maximal subgroups of orthogonal groups},
@@ -43,11 +39,12 @@ MRREVIEWER = {A. S. Kondrat\cprime ev},
     VOLUME = {13},
       YEAR = {2010},
      PAGES = {164--191},
-   MRCLASS = {20D06 (20E28 20G40 68Q25)},
+   MRCLASS = {20H30 (20D06 20E28 20G40 68Q25)},
+  MRNUMBER = {2645140},
+MRREVIEWER = {Mohammad-Reza Darafsheh},
        DOI = {10.1112/S1461157009000035},
        URL = {https://doi.org/10.1112/S1461157009000035},
 }
-
 
 @book {KL90,
     AUTHOR = {Kleidman, Peter and Liebeck, Martin},
--- a/doc/ClassicalMaximals.bib
+++ b/doc/ClassicalMaximals.bib
@@ -31,6 +31,24 @@ MRREVIEWER = {A. S. Kondrat\cprime ev},
        URL = {https://doi.org/10.1112/S1461157000000899},
 }
 
+%%%%%%%%%%%%%%%%%%%%%%%%%%%
+% TODO
+% Add MRNUMBER, MRREVIEWER
+%%%%%%%%%%%%%%%%%%%%%%%%%%%
+@article {HR10,
+    AUTHOR = {Holt, Derek F. and Roney-Dougal, Colva M.},
+     TITLE = {Constructing maximal subgroups of orthogonal groups},
+   JOURNAL = {LMS J. Comput. Math.},
+  FJOURNAL = {LMS Journal of Computation and Mathematics},
+    VOLUME = {13},
+      YEAR = {2010},
+     PAGES = {164--191},
+   MRCLASS = {20D06 (20E28 20G40 68Q25)},
+       DOI = {10.1112/S1461157009000035},
+       URL = {https://doi.org/10.1112/S1461157009000035},
+}
+
+
 @book {KL90,
     AUTHOR = {Kleidman, Peter and Liebeck, Martin},
      TITLE = {The subgroup structure of the finite classical groups},
--- a/gap/Forms.gi
+++ b/gap/Forms.gi
@@ -5,14 +5,19 @@
 InstallGlobalFunction("ConjugateToSesquilinearForm",
 function(group, type, gramMatrix)
     local gapForm, newForm, gapToCanonical, canonicalToNew, field, formMatrix,
-        result, d, q;
+        result, d, q, broadType;
     if not type in ["S", "O-B", "O-Q", "U"] then
-        ErrorNoReturn("<type> must be one of 'S', 'U', 'O'");
+        ErrorNoReturn("<type> must be one of 'S', 'U', 'O-B', 'O-Q'");
     fi;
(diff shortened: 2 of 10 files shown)


maxhauck:
@fingolfin After being a bit puzzled for a while, we discovered that GAP uses different standard orthogonal forms for `SO(epsilon, d, q)` and `Omega(epsilon, d, q)` ... - we therefore decided to conjugate the constructed subgroup so that it is inside `Omega(epsilon, d, q)` and consequently not inside `SO(epsilon, d, q)`, which should be what we want, right?

fingolfin:
Regarding Omega: this is known, see https://github.com/gap-system/gap/issues/500. Since it's been like this "forever", we can't simply change it without risking to break existing code out there. The solution we came up with is that in GAP 4.12, various constructors will allow the user to *optionally* specify a form.

gap/ReducibleMatrixGroups.gi
```diff
@@ -486,3 +486,124 @@ function(d, q, k)
...
+                             field);
+    else
+        gamma := FindGamma(q);
+        F := AntidiagonalMat(d, field);
+        Q := AntidiagonalMat(Concatenation(ListWithIdenticalEntries(m, one),
+                                           ListWithIdenticalEntries(m, zero)),
+                             field);
+        Q[m, m] := one;
+        Q[m + 1, m + 1] := gamma;
+    fi;
```
fingolfin, line 525:
```suggestion
    F := AntidiagonalMat(d, field);
    Q := AntidiagonalMat(Concatenation(ListWithIdenticalEntries(m, one),
                                       ListWithIdenticalEntries(m, zero)),
                         field);
    if epsilon = -1 then
        gamma := FindGamma(q);
        Q[m, m] := one;
        Q[m + 1, m + 1] := gamma;
    fi;
```

gap/Utils.gi
```diff
@@ -136,6 +136,67 @@ function(type, alpha, q)
...
+    # We have (a / b ^ 2) * (a * x ^ 2 + b * x + c) 
+    #       = (a / b * x) ^ 2 + (a / b * x) + (c * a / b ^ 2) 
+    # Hence we find a solution of t ^ 2 + t + c * a / b ^ 2 = 0.
+    d := c * a / b ^ 2; 
+
+    # Note that the map t --> t ^ 2 + t is linear so we can express it via a
+    # representation matrix and find a pre-image of -d (if one exists) by
+    # solving a linear system 
+    V := AsVectorSpace(primeField, F);
+    B := Basis(V);
```
fingolfin, line 191: You don't need `V` nor `primeField`:
```suggestion
    B := CanonicalBasis(F);
```

gap/Utils.gi
```diff
@@ -136,6 +136,67 @@ function(type, alpha, q)
...
+    M := List([1..e], i -> Coefficients(B, B[i] + B[i] ^ 2));
```
fingolfin, line 192:
```suggestion
    M := List(B, b -> Coefficients(B, b + b ^ 2));
```

gap/Utils.gi
```diff
@@ -136,6 +136,67 @@ function(type, alpha, q)
...
+    F := GF(q);
+    R := PolynomialRing(F, ["x"]);
+    x := Indeterminate(F, "x");
+
+    for gamma in F do
+        polynomial := x ^ 2 + x + gamma;
+        if IsIrreducibleRingElement(R, polynomial) then
+            return gamma;
+        fi;
+    od;
```
fingolfin, line 160: That's a brute force search which seems inefficient. Can't  you adapt the idea from `SolveQuadraticEquation`: the map `t -> t^2 + t` is linear in char 2. You are looking for value `gamma` that is *not* in its image. So like this:

```gap
FindGamma2 := function(q)
    local F, B, M, i;

    F := GF(q);
    B := CanonicalBasis(F);
    M := List(B, b -> Coefficients(B, b + b ^ 2));
    ConvertToMatrixRep(M);
    M := RREF(M);
    i := First([1..Length(M)], i -> IsZero(M[i,i]));
    return B[i];
end;
```
For q=2^17 the original `FindGamma` runs a long time for me, because there is no conway polynomial and no good polynomial factoring is available in that field in GAP; while `FindGamma2` is instant.

maxhauck: Okay, this makes sense: It's not the actual brute force search that's inefficient (because half of all possible `gamma` make `x ^ 2 + x + gamma` irreducible by a simple linear algebra argument), but the test for irreducibility involved in the process! ;)

doc/ClassicalMaximals.bib
```diff
@@ -31,6 +31,24 @@ @article {HR05
...
+  FJOURNAL = {LMS Journal of Computation and Mathematics},
+    VOLUME = {13},
+      YEAR = {2010},
+     PAGES = {164--191},
+   MRCLASS = {20D06 (20E28 20G40 68Q25)},
+       DOI = {10.1112/S1461157009000035},
+       URL = {https://doi.org/10.1112/S1461157009000035},
+}
+
+
```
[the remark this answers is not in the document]
maxhauck: Not that important indeed, but since you added this information for all the other bibliography entries as well, it's much nicer this way in terms of consistency :)

gap/Forms.gi
```diff
@@ -113,26 +129,34 @@ function(group, type)
...
-    fi;
     
     # get standard GAP form
     if type = "S" then
         gapForm := InvariantBilinearForm(Sp(d, q)).matrix;
     elif type = "U" then
         gapForm := InvariantSesquilinearForm(GU(d, q)).matrix;
     elif type = "O" then
-        gapForm := InvariantBilinearForm(GO(d, q)).matrix;
+        gapForm := InvariantBilinearForm(Omega(d, q)).matrix;
```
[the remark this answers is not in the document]
maxhauck: Ad 1st comment. I would also favour option 2, but with one extra bit: We could simply agree on (and this would actually make much sense) always returning the group so that it preserves the same form as the "standard" version of the corresponding quasisimple group in GAP. Why is this senseful? Well, we actually want to construct subgroups _of these quasisimple groups_, so one would expect that they are least contained in the corresponding quasisimple group provided by GAP. (Using the form of the corresponding quasisimple group in GAP as a default is also a simple way of explaining what the `ConjugateToStandardForm` function does.) - Let's talk about this tomorrow though!

Ad 2nd comment. I guess this would actually not really work and for the following reason: It is also possible to use a group _without_ having none of the parameters `InvariantXForm` as an input for the function `ConjugateToStandardForm` - in this case, the invariant form is determined (at least for the unitary, symplectic and orthogonal odd char cases - yes, we should add orthogonal char 2 here) in this last line ``return ConjugateToSesquilinearForm(group, broadType, gapForm)``. In particular, the function `ConjugateToSesquilinearForm` needs to be told for which type of form to look so we cannot simply omit the argument `type` in these cases.

fingolfin (approved):

PR #88, opened 2021-12-10
Set invariant forms instead of using the MeatAxe

maxhauck:
I have discovered how to directly compute the forms preserved by the C3, C4 and C7 subgroups we construct instead of determining them via the MeatAxe! - Somehow, a RECOG error has crept up, which is the only reason why the tests are failing ... I'll deal with that shortly.

### General

### DIFF
--- a/gap/Forms.gi
+++ b/gap/Forms.gi
@@ -186,7 +186,7 @@ function(i, dims)
         s := t - s;
         r := q mod dims[s];
         result := Concatenation([r + 1], result);
-        q := (q - r) / dims[s];
+        q := QuoInt(q, dims[s]);
     od;
 
     return result;
--- a/gap/Forms.gi
+++ b/gap/Forms.gi
@@ -164,6 +164,61 @@ function(group, type)
     return ConjugateToSesquilinearForm(group, broadType, gapForm);
 end);
 
+# Let <dims> = [d1, d2, ..., dt], consider the tensor product of the vector
+# spaces F ^ d1, F ^ d2, ..., F ^ dt and in it the i-th basis vector ei. 
+#
+# Return a list of indices [i1, i2, ..., it] such that ei is the tensor product 
+# of the basis vectors ei1, ei2, ..., eit of F ^ d1, F ^ d2, ..., F ^ dt.
+BindGlobal("GetTensorFactors",
+function(i, dims)
+    local d, t, q, s, r, result;
+
+    d := Product(dims);
+    t := Length(dims);
+    if i > d or i < 1 then
+        ErrorNoReturn("<i> must be between 0 and the product of all elements",
+                      " of <dims>");
+    fi;
+
+    result := [];
+    q := i - 1;
+    for s in [0..t - 1] do
+        s := t - s;
+        r := q mod dims[s];
+        result := Concatenation([r + 1], result);
+        q := (q - r) / dims[s];
+    od;
+
+    return result;
+end);
+
+# Let <forms> = [f1, f2, ..., ft] be a list of sesquilinear forms on the vector
+# spaces F ^ d1, F ^ d2, ..., F ^ dt. Then we can lift these to a sesquilinear
+# form f on the tensor product F ^ d1 x F ^ d2 x ... x F ^ dt by defining
+# f(v1 x v2 x ... x vt, w1 x w2 x ... x wt) = f1(v1, w1)f2(v2, w2)...ft(vt, wt).
+#
+# Return the Gram matrix of f; the forms f1, f2, ..., ft must be given as their
+# respective Gram matrices.
+BindGlobal("LiftFormsToTensorProduct",
+function(forms, F)
+    local dims, d, t, newForm, i, j, indicesi, indicesj;
+
+    dims := List(forms, f -> Size(f));
+    d := Product(dims);
+    t := Length(dims);
+    newForm := NullMat(d, d, F);
+
+    for i in [1..d] do
+        for j in [1..d] do
+            indicesi := GetTensorFactors(i, dims);
+            indicesj := GetTensorFactors(j, dims);
+
+            newForm[i, j] := Product(List([1..t], k -> (forms[k])[indicesi[k], indicesj[k]]));
+        od;
+    od;
+
+    return newForm;
+end);
 
 BindGlobal("ConjugateModule",
 function(M, q)
--- a/gap/Forms.gi
+++ b/gap/Forms.gi
@@ -203,17 +203,17 @@ BindGlobal("LiftFormsToTensorProduct",
 function(forms, F)
     local dims, d, t, newForm, i, j, indicesi, indicesj;
 
-    dims := List(forms, f -> Size(f));
+    dims := List(forms, NrRows);
     d := Product(dims);
     t := Length(dims);
     newForm := NullMat(d, d, F);
 
     for i in [1..d] do
+        indicesi := GetTensorFactors(i, dims);
         for j in [1..d] do
-            indicesi := GetTensorFactors(i, dims);
             indicesj := GetTensorFactors(j, dims);
 
-            newForm[i, j] := Product(List([1..t], k -> (forms[k])[indicesi[k], indicesj[k]]));
+            newForm[i, j] := Product([1..t], k -> (forms[k])[indicesi[k], indicesj[k]]);
         od;
     od;
 
--- a/gap/SemilinearMatrixGroups.gi
+++ b/gap/SemilinearMatrixGroups.gi
@@ -139,7 +139,8 @@ end);
 BindGlobal("GammaLMeetSU",
 function(d, q, s)
     local F, As, Bs, Cs, Fs, m, gammaL1, Y, AandB, C, D, i,
-    range, AsInGU, omega, generators, size;
+    range, AsInGU, omega, generators, size, result, standardForm, formMatrix,
+    j, u, v, B;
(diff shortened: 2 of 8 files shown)


gap/Utils.gi
```diff
@@ -194,6 +225,14 @@ function(F, a, b, c)
     return b / a * t;
 end);
 
+# Compute the Trace of an element x in GF(q ^ s) over GF(q). Since the Galois
+# group of GF(q ^ s) over GF(q) is cyclic with generator x -> x ^ q, this is
+# just x + x ^ q + x ^ (q  ^ 2) + ... + x ^ (q  ^ (s - 1))
+BindGlobal("TraceOverFiniteField",
+function(x, q, s)
+    return Sum(List([0..s - 1], i -> x ^ (q ^ i)));
+end);
```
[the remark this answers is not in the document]
maxhauck: Well, I should have guessed that there already exists a function for this ... However, now it's there and for the special case of finite fields, this custom version might even have a slightly better performance, so I'll just leave this as is.

gap/Utils.gi
```diff
@@ -66,6 +66,37 @@ function(M, func)
     return result;
 end);
 
+# Take an m x n-matrix M and change its shape to a rows x cols-matrix, where
+# entries are copied column by column. 
+BindGlobal("ReshapeMatrix",
```
fingolfin, line 71: This is not used anymore... THough perhaps it'll be handy elsewhere later on?

fingolfin (approved):

fingolfin:
Unfortunately CI tests fail:
```
TestTensorProductStabilizerInSp([1, 2, 4, 5]);
# Expected output:
# But found:
Error, List Element: <list>[3] must have an assigned value
```

gap/Utils.gi
maxhauck: Perhaps - I guess it doesn't do much harm so ...

maxhauck:
> Unfortunately CI tests fail:
> 
> ```
> TestTensorProductStabilizerInSp([1, 2, 4, 5]);
> # Expected output:
> # But found:
> Error, List Element: <list>[3] must have an assigned value
> ```

I removed the corresponding test and noted down a random seed for which this fails.

gap/Forms.gi
```diff
@@ -164,6 +164,61 @@ function(group, type)
     return ConjugateToSesquilinearForm(group, broadType, gapForm);
 end);
 
+# Let <dims> = [d1, d2, ..., dt], consider the tensor product of the vector
+# spaces F ^ d1, F ^ d2, ..., F ^ dt and in it the i-th basis vector ei. 
+#
+# Return a list of indices [i1, i2, ..., it] such that ei is the tensor product 
+# of the basis vectors ei1, ei2, ..., eit of F ^ d1, F ^ d2, ..., F ^ dt.
```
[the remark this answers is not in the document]
fingolfin: However, after writing this, and thus finally understanding what it does, I am not even convinced it should be kept. I think it would be more efficient and easier to maintain if the code below was changed to use `IteratorOfCartesianProduct(List(dims,d->[1..d]))`.

Alas, I don't want to require either of you here; such a change could be made later, if at all.

gap/SemilinearMatrixGroups.gi
```diff
@@ -137,11 +137,86 @@ function(n, q, s)
...
 
+# For a sesquilinear form beta on the vector space GF(q ^ s) ^ (d / s) given 
+# by the Gram matrix <form>, we can interpret any u, v in GF(q) ^ d as elements of 
+# GF(q ^ s) ^ (d / s) via the standard basis. Taking the trace of beta(u, v)
+# yields a sesquilinear form on the vector space GF(q) ^ d.
+#
+# Takes a unitary form and returns a unitary form if <type> = "U-U"; takes a
+# unitary form and returns a symplectic form if <type> = "U-S"; takes a
+# symplectic form and returns a symplectic form if <type> = "S-S". 
+BindGlobal("TakeTraceOfSesquilinearForm",
```
fingolfin, line 148: Is this really a good name? Perhaps something like `ReinterpretSesquilinearForm` or `LiftSesquilinearForm` would be more appropriate?

gap/SemilinearMatrixGroups.gi
```diff
@@ -137,11 +137,86 @@ function(n, q, s)
...
+                newForm[i, j] := TraceOverFiniteField(eiOverGFqsEntry
+                                                            * form[eiOverGFqsIndex, ejOverGFqsIndex] 
+                                                            * ejOverGFqsEntry ^ q, 
+                                                      q, s);
+            else
+                newForm[i, j] := TraceOverFiniteField(eiOverGFqsEntry 
+                                                            * form[eiOverGFqsIndex, ejOverGFqsIndex]
+                                                            * ejOverGFqsEntry ^ (rootOfq ^ s), 
+                                                      q, s);
+            fi;
```
fingolfin, line 198: How about adding a local variable `exp` which is set to `1`, `q` or `rootOfq ^ s` at the start, depending on `type`; and another local variable `tmp`; and then this code can be simplified to the following IMHO more readable variant
```suggestion
            tmp := eiOverGFqsEntry * form[eiOverGFqsIndex, ejOverGFqsIndex] * ejOverGFqsEntry ^ exp;
            newForm[i, j] := TraceOverFiniteField(tmp, q, s);
```

tst/quick/Utils.tst
```diff
@@ -11,6 +11,9 @@ gap> M := GL(7, 5 ^ 2).1;;
 gap> N := ApplyFunctionToEntries(M, x -> x ^ 5);;
 gap> M = ApplyFunctionToEntries(N, x -> x ^ 5);
 true
+gap> M := IdentityMat(12, GF(7));;
+gap> Size(ReshapeMatrix(M, 48, 3)) = 48;
```
fingolfin, line 15: Don't call `Size` on a matrix, use `NrRows`; or here, probably `DimensionsMat` would be best. Well... and/or just use a smaller example so we can actually preview it...

```suggestion
gap> M := IdentityMat(4);;
gap> ReshapeMatrix(M, 2, 8);
[ [ 1, 0, 0, 0, 0, 1, 0, 0 ], [ 0, 0, 1, 0, 0, 0, 0, 1 ] ]
gap> M := IdentityMat(12, GF(7));;
gap> DimensionsMat(ReshapeMatrix(M, 48, 3));
[ 48, 3 ]
```

fingolfin (approved):
Could be merged, but I still have some nitpicks, in case you are willing...

gap/Forms.gi
```diff
@@ -164,6 +164,61 @@ function(group, type)
...
+        s := t - s;
+        r := q mod dims[s];
+        result := Concatenation([r + 1], result);
+        q := QuoInt(q, dims[s]);
+    od;
+
+    return result;
+end);
+
+# Let <forms> = [f1, f2, ..., ft] be a list of sesquilinear forms on the vector
```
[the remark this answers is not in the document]
maxhauck: It literally says that the forms must be given as Gram matrices in the paragraph after that, but if you think additional clarification is needed then go for it ...

gap/SemilinearMatrixGroups.gi
maxhauck: Well, I'd say it is definitely not lifting since we take a sesquilinear form on a vector space over a bigger field `GF(q ^ s)` and turn it into a sesquilinear form on a vector space over a smaller field `GF(q)`.

I thought the name was really quite fitting since if `f` is a sesquilinear form on `GF(q ^ s) ^ (d / s)`, then the function returns a sesquilinear form `f'` on `GF(q) ^ d` satisfying `f'(u, v) = Tr(f(u, v))`, i.e. one literally takes the trace of the value of the sesquilinear form `f` to get the value of `f'`.

gap/Forms.gi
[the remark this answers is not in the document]
maxhauck: I think the choice is not as arbitrary as it might seem: I believed it to be quite common to mean by `e_i` the `i`-th standard basis vector (i.e. with a `1` in the `i`-th component and zeroes elsewhere). Then each `e_i` can be uniquely written as some *Kronecker product* `e_i1 x e_i2 x ... x e_it` and the function returns exactly the indices `i1, ..., it`. 

So I guess the wording of the comment should be changed slightly so that it states we are assuming `e_i` to be the `i`-th standard basis vector and that we "choose" the tensor product of a collection of vectors to be their Kronecker product?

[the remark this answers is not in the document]
fingolfin: But you are referring with ei to two different things: one is the i-th standard basis vector in K^n, and I agree this is common enough. But then in line 170 you use it to refer to a basis vector in the Kronecker product, and there this is *not* clear. My main point is that the *order* of tuples is arbitrary, and should be documented

gap/SemilinearMatrixGroups.gi
fingolfin: Well, it doesn't return `Tr(f(u,v))` though, it returns `Tr\circ f`, so that would suggest `ComposeTraceAndForm` would fit that. But even if we stay with your argument: then why "Take" and not just "TraceOfSesquilinearForm" ???

Anyway, this is not important at all for the PR overall, and can be changed later, if at all.

fingolfin (approved):

fingolfin:
Thank you!

gap/Forms.gi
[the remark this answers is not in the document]
fingolfin: I left it as-is since I propose to replace this code anyway, see PR #95

PR #89, opened 2021-12-10
Symplectic Maximals

TristanPfersdorff:
- Add `MaximalSubgroupClassRepsSpecialUnitaryGroup`
- Modify `ClassicalMaximalsGeneric`
- Various fixes for classes C2, C5, C6, C7
- Test `MaximalSubgroupClassRepsSpecialUnitaryGroup`

With this PR, the maximal geometric subgroups of the symplectic groups should be all done.

### General
- [ ] All functions have unit tests

### Functions constructing generators of maximal subgroups
- [ ] The code which computes and then stores the sizes is correct. To do this one confers to the tables referenced by the code. Tests for the group size should use `RECOG.TestGroup` from the recog package.
- [ ] The test suite checks whether all constructed generators are sensible. That is it tests the generators'
  - [ ] determinants (and if applicable also the spinor norms are correct), and
  - [ ] compatibility with the correct form,
  - [ ] `DefaultFieldOfMatrixGroup` returns the correct field.

### Functions assembling the list of all maximal subgroups of a certain group
- [ ] The code agrees with the tables in section 8.2 of the book "The Maximal Subgroups of the Low-dimensional Finite Classical Groups".

The reviewer doesn't need to compare our results to magma's results. That's the job of the person implementing the code.

PR #91, opened 2021-12-17
Revise tests to check forms directly

fingolfin:
Testing `IsSubset(G, elm)` can be slow in GAP if G is a matrix group of
"large" degree. This is because GAP resorts to testing membership using a
stabilizer chain in an isomorphic permutation group, even if G is a
linear/unitary/orthogonal/symplectic group. So we replace that check with one
that actually tests using the invariant forms of the groups in question.

This is work in progress, and <s>waiting for PR #88</s> and needs a few functions such as `SpIsotropicImprimitives` to set invariant forms -- which PR #89 seems to do.

### General

### DIFF
--- a/gap/SubfieldMatrixGroups.gi
+++ b/gap/SubfieldMatrixGroups.gi
@@ -307,27 +307,28 @@ function (d, p, e, f)
     field := GF(p ^ e);
     q0 := p ^ f;
     gens := List(GeneratorsOfGroup(Sp(d, q0)));
+    l := QuoInt(d, 2);
 
-    # In this case the embedding of Sp(d, q0) in Sp(d, q) is already
-    # the C5-subgroup, so we just need to adjust the base field.
     if Gcd(2, b, p - 1) = 1 then
-        return MatrixGroupWithSize(field, gens, SizeSp(d, q0));
+        # In this case the embedding of Sp(d, q0) in Sp(d, q) is already
+        # the C5-subgroup, so we just need to adjust the base field.
+        result := MatrixGroupWithSize(field, gens, SizeSp(d, q0));
+    else
+        zeta := PrimitiveElement(field);
+        omega := PrimitiveElement(GF(q0));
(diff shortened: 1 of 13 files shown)


fingolfin:
@TristanPfersdorff these tests caught one omission in `SubfieldSp` (it did not always set the invariant form, fixed in this PR), and potentially also a "real" bug in `SymplecticSubfieldSU`: for inputs `n=2,q=4` the returned group does not preserve the given sesquilinear form. Indeed, according to a brute force search, it doesn't seem to preserve any such form:
```
gap> G:=SymplecticSubfieldSU(2,4);;
gap> Filtered(GF(4)^[2,2], M -> ForAll(GeneratorsOfGroup(G), g -> g*M*HermitianConjugate(g,2) = M));
[ [ [ 0*Z(2), 0*Z(2) ], [ 0*Z(2), 0*Z(2) ] ] ]
```

fingolfin:
Oh no wait, I misunderstood the function arguments, the base field is `GF(4)` and the big field is `GF(16)`. OK, so trying again:
```
gap> G:=SymplecticSubfieldSU(2,4);;
gap> forms:=Filtered(GF(4)^[2,2], M -> ForAll(GeneratorsOfGroup(G), g -> g*M*HermitianConjugate(g,4) = M));
[ [ [ 0*Z(2), 0*Z(2) ], [ 0*Z(2), 0*Z(2) ] ], [ [ 0*Z(2), Z(2)^0 ], [ Z(2)^0, 0*Z(2) ] ],
  [ [ 0*Z(2), Z(2^2) ], [ Z(2^2), 0*Z(2) ] ], [ [ 0*Z(2), Z(2^2)^2 ], [ Z(2^2)^2, 0*Z(2) ] ] ]
gap> Display(forms[2]);
 . 1
 1 .
gap> Display(InvariantSesquilinearForm(G).matrix);
 . 1
 1 .
gap> CheckSesquilinearForm(G);
false
```

The bug was in `CheckSesquilinearForm`, fixed now.

PR #93, opened 2021-12-22
More test tweaks

fingolfin:
- Rewrite tst/standard/ClassicalMaximals.tst
- tests: avoid needless use of variadic functions


ssiccha (approved):
LGTM, I only roughly scanned the diff though.

PR #100, opened 2022-01-09
Orthogonal C1 NonDegenerate

TristanPfersdorff:
Adds the construction for the stabilizer of the canonical non-degenerate-subspace according to Lemma 4.3 in [HR10] via the function `OmegaStabilizerOfNonDegenerateSubspace` along with some error handling tests for `StandardGeneratorsOfOrthogonalGroup` and `StandardOrthogonalForm` as well as a minor addition to `StandardGeneratorsOfOrthogonalGroup`, see my comment below for details.

### General

### DIFF
--- a/tst/standard/ReducibleMatrixGroups.tst
+++ b/tst/standard/ReducibleMatrixGroups.tst
@@ -75,6 +75,25 @@ gap> SpStabilizerOfNonDegenerateSubspace(4, 2, 3);
 Error, <k> must be less than <d> / 2
 
 #
+gap> TestOmegaStabilizerOfNonDegenerateSubspace := function(epsilon, d, q, epsilon_0, k)
+>   local G;
+>   G := OmegaStabilizerOfNonDegenerateSubspace(epsilon, d, q, epsilon_0, k);
+>   Assert(0, CheckSize(G));
+>   Assert(0, IsSubset(Omega(epsilon, d, q), GeneratorsOfGroup(G)));
+>   Assert(0, DefaultFieldOfMatrixGroup(G) = GF(q));
+> end;;
+gap> TestOmegaStabilizerOfNonDegenerateSubspace(0, 7, 5, 1, 3);
+gap> TestOmegaStabilizerOfNonDegenerateSubspace(0, 7, 5, -1, 5);
+gap> TestOmegaStabilizerOfNonDegenerateSubspace(1, 8, 5, -1, 2);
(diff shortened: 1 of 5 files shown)


TristanPfersdorff:
I had to add the special case `d = 1` to `StandardGeneratorsOfOrthogonalGroup`, since otherwise a few statements involving `d - 2` cause errors. It feels a little hacky but is mathematically justified. The field `generatorsOfOmega` could also be an empty list in that case (GAP even does this by default, calling `GeneratorsOfGroup` on a trivial group returns an empty list), but this would cause some issues in `OmegaStabilizerOfNonDegenerateSubspace` and probably also future construction functions, so my current solution of returning just the identity matrix as a generator seems like a good choice, but this is subject to review.
In other news, one of the tests causes a weird `Error, !!!`, similar (but not necessarily related) to [this RECOG-issue](https://github.com/gap-packages/recog/issues/12).

tst/standard/ReducibleMatrixGroups.tst
```diff
@@ -74,6 +74,63 @@ Error, <d> must be even
 gap> SpStabilizerOfNonDegenerateSubspace(4, 2, 3);
 Error, <k> must be less than <d> / 2
 
+#
+gap> TestOmegaStabilizerOfNonDegenerateSubspace := function(epsilon, d, q, epsilon_0, k)
+>   local G;
+>   G := OmegaStabilizerOfNonDegenerateSubspace(epsilon, d, q, epsilon_0, k);
+>   Assert(0, CheckSize(G));
+>   Assert(0, IsSubset(Omega(epsilon, d, q), GeneratorsOfGroup(G)));
```
fingolfin, line 82: It'd be preferable to add a function `IsSubsetOmega` in `tst/utils.g`. It could for now look like this:
```
IsSubsetSU := function(epsilon, n, q, G)
  Assert(0, DimensionOfMatrixGroup(G) = n);
  Assert(0, DefaultFieldOfMatrixGroup(G) = GF(q));
  Assert(0, CheckGeneratorsSpecial(G));
  Assert(0, CheckQuadraticForm(G));
  return true;
end;
```
But of course this does not yet verify `epsilon`, nor that we are really in `Omega`.... I guess for that we'll want `SpinorNorm` from https://github.com/gap-system/gap/pull/4668 (so only in GAP 4.12). Err, and yet more code in char 2... *sigh*. Certainly nothing to delay this PR over, but we should track this in an issue.

fingolfin (approved):

PR #102, opened 2022-01-15
Orthogonal `IsSubsetOmega`

### LINKED ISSUE #101: Write `IsSubsetOmega` and refactor tests accordingly
We could really use a function `IsSubsetOmega` similar to the other functions in `tst/utils.g` to test the orthogonal construction functions. This is not completely trivial as outlined [here](https://github.com/gap-packages/ClassicalMaximals/pull/100/files#r784802325).

TristanPfersdorff:
Fixes #101.

### General

### DIFF
--- a/tst/utils.g
+++ b/tst/utils.g
@@ -66,3 +66,27 @@ IsSubsetSU := function(n, q, G)
   Assert(0, CheckSesquilinearForm(G));
   return true;
 end;
+
+IsSubsetOmega := function(epsilon, n, q, G)
+  local field, invariantForm;
+  field := GF(q);
+  if IsEvenInt(q) then
+    Assert(0, CheckQuadraticForm(G));
+    invariantForm := QuadraticFormByMatrix(InvariantQuadraticForm(G).matrix, field);
+  else
+    Assert(0, CheckBilinearForm(G));
+    invariantForm := BilinearFormByMatrix(InvariantBilinearForm(G).matrix, field);
+  fi;
+  if epsilon = 0 then
+    Assert(0, WittIndex(invariantForm) = QuoInt(n - 1, 2));
+  elif epsilon = 1 then
+    Assert(0, WittIndex(invariantForm) = QuoInt(n, 2));
+  elif epsilon = -1 then
+    Assert(0, WittIndex(invariantForm) = QuoInt(n, 2) - 1);
+  fi;
+  Assert(0, DimensionOfMatrixGroup(G) = n);
+  Assert(0, DefaultFieldOfMatrixGroup(G) = field);
+  Assert(0, CheckGeneratorsSpecial(G));
+  Assert(0, ForAll(GeneratorsOfGroup(G), g -> FancySpinorNorm(InvariantBilinearForm(G).matrix, field, g) = 1));
+  return true;
+end;
--- a/tst/utils.g
+++ b/tst/utils.g
@@ -70,6 +70,9 @@ end;
 IsSubsetOmega := function(epsilon, n, q, G)
   local field, invariantForm;
   field := GF(q);
+  Assert(0, DimensionOfMatrixGroup(G) = n);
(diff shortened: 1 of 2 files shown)


TristanPfersdorff:
I did some digging and the best way I found to test `epsilon` is by checking the Witt Index, which is possible with the function `WittIndex` from the forms package. I used our custom `FancySpinorNorm` to check the spinor norm.
The function looks a little bit chaotic since we have to differentiate between `q` odd and `q` even, because `OmegaStabilizerOfNonDegenerateSubspace` sets the group's quadratic form if `q` is even and its bilinear form otherwise. Setting the quadratic form also automatically sets the bilinear form, but not the other way around, even though that would be possibe if `q` is odd. What I'm really asking is: Should we set the quadratic form instead of (or in addition to) the bilinear form in case `q` odd in the future, just to make sure the attribute is always set?

maxhauck:
> Should we set the quadratic form instead of (or in addition to) the bilinear form in case `q` odd in the future, just to make sure the attribute is always set?

To me, that sounds like a very sensible thing to do 👍

fingolfin:
Agreed

tst/utils.g
```diff
@@ -66,3 +66,27 @@ IsSubsetSU := function(n, q, G)
...
+    Assert(0, CheckBilinearForm(G));
+    invariantForm := BilinearFormByMatrix(InvariantBilinearForm(G).matrix, field);
+  fi;
+  if epsilon = 0 then
+    Assert(0, WittIndex(invariantForm) = QuoInt(n - 1, 2));
+  elif epsilon = 1 then
+    Assert(0, WittIndex(invariantForm) = QuoInt(n, 2));
+  elif epsilon = -1 then
+    Assert(0, WittIndex(invariantForm) = QuoInt(n, 2) - 1);
+  fi;
```
fingolfin, line 86: Could also do
```suggestion
  Assert(0, WittIndex(invariantForm) = QuoInt(n - 1 + epsilon, 2));
```
or
```suggestion
  Assert(0, 2 * WittIndex(invariantForm) = n + epsilon - 1);
```
or so. Not that it matters either way :-).

tst/utils.g
```diff
@@ -66,3 +66,27 @@ IsSubsetSU := function(n, q, G)
...
+  Assert(0, DimensionOfMatrixGroup(G) = n);
+  Assert(0, DefaultFieldOfMatrixGroup(G) = field);
+  Assert(0, CheckGeneratorsSpecial(G));
```
fingolfin, line 89: In the other functions I started with these three, as they are sorted in order of "price", i.e., the cheapest check first, etc. -- there's no point in checking the spinor norm if the matrices have the wrong dimension or are not invertible. In the latter case, that ensures slightly more helpful errors, too (e.g. don't waste time debugging the Witt index when the real issue is that the matrix was not invertible)

But again, it doesn't really matter

fingolfin (approved):
Thanks! Could be merged, or you could address my minor remarks, your choice. Also, I like the suggestion to always set both bilinear and quadradtic form (where possible/sensible, at least) and then simplifying the test function here.

TristanPfersdorff:
The suggestions seemed quite nice to me, especially the order of assertions: I was thinking about the order and I concluded that it doesn't matter, since I only had performance in mind and the performance only differs if the tests fail -- but the order/priority in which error messages appear is quite an important argument, and I agree that the more serious/obvious errors like an incorrect matrix size should always be called out first.
As for the whole "just always set the quadratic form", I thought that would be quite easy and I was mistaken: The current functions we use for conjugating everything to GAP's standard form are quite reliant on using bilinear/sesquilinear forms; in particularly, they only set the bilinear form after conjugation and not the quadratic form, so we'd have to change that. Which is entirely possible (and probably not that much effort), but a little more than I bargained for with this PR. We should probably take a look at that issue separately, since all our conjugation-infrastructure is quite chaotic when it comes to orthogonal groups.

fingolfin (approved):

PR #103, opened 2022-01-17
Implement standard generators of groups S and G in cases L, S, U

TristanPfersdorff:
Implement standard generators of groups S and G in cases L, S, U with the requirements of Theorem 3.11 in [HR10]. This is needed for future constructions in the O case.
The constructions are given in [T87], which I was not able to cite very well, presumably since according to [Taylor's website](https://www.maths.usyd.edu.au/u/don/publications.html), it is only a technical report and not a research paper (I can't seem to find any fitting URL or doi, I also cannot find it in the library of the TU Kaiserslautern or of the University of Sydney).
This report also does not seem to fit Theorem 3.11 in [HR10] perfectly, so I'm taking some creative liberties to ensure that the generators are suitable for future constructions.

### General

### DIFF
--- a/doc/ClassicalMaximals.bib
+++ b/doc/ClassicalMaximals.bib
@@ -92,7 +92,10 @@ MRREVIEWER = {Irina Suprunenko},
 
 @article {T87,
     AUTHOR = {Taylor, Donald E.},
-     TITLE = {Conjugacy of subgroups of the general linear group},
+     TITLE = {Pairs of Generators for Matrix Groups, I},
+   JOURNAL = {The Cayley Bulletin},
+    VOLUME = {3},
       YEAR = {1987},
+       URL = {https://www.maths.usyd.edu.au/u/don/papers/genAC.pdf},
 }
 
--- a/doc/ClassicalMaximals.bib
+++ b/doc/ClassicalMaximals.bib
@@ -90,3 +90,9 @@ MRREVIEWER = {Irina Suprunenko},
        URL = {https://doi.org/10.1016/j.jsc.2010.09.003},
 }
 
+@article {T87,
+    AUTHOR = {Taylor, Donald E.},
+     TITLE = {Conjugacy of subgroups of the general linear group},
+      YEAR = {1987},
+}
+
--- a/gap/Utils.gi
+++ b/gap/Utils.gi
@@ -622,12 +622,12 @@ function(d, q)
 
     elif d = 2 and q = 3 then
 
-        L1 := [[-one, one], [-one, 0 * one]];
-        L2 := [[-one, one], [one, 0 * one]];
+        L1 := one * [[-1, 1], [-1, 0]];
+        L2 := one * [[-1, 1], [1, 0]];
 
         # This is precisely L2^2, which is how we ensure that
         # L1 and L2^2 = L3 generate (1 / 2) GL(2, 3) = SL(2, 3).
-        L3 := [[-one, -one], [-one, one]];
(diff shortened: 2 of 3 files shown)


TristanPfersdorff:
On second thought, I really don't think we need to implement any of the standard generators for the S and U case. Their only special properties according to Theorem 3.11 would be that all possible entries of the matrices are known, and I really don't think we actually care about that. With that in mind, this PR is probably ready, the title is not quite apt anymore.

doc/ClassicalMaximals.bib
```diff
@@ -90,3 +90,9 @@ @article {MR11
        URL = {https://doi.org/10.1016/j.jsc.2010.09.003},
 }
 
+@article {T87,
+    AUTHOR = {Taylor, Donald E.},
+     TITLE = {Conjugacy of subgroups of the general linear group},
```
fingolfin, line 95: Don't you mean
```suggestion
     TITLE = {Pairs of Generators for Matrix Groups, I},
```

doc/ClassicalMaximals.bib
```diff
@@ -90,3 +90,9 @@ @article {MR11
...
+      YEAR = {1987},
```
fingolfin, line 96:
```suggestion
      JOURNAL = {The Cayley Bulletin},
      VOLUME = {3},
      YEAR = {1987},
      URL = {https://www.maths.usyd.edu.au/u/don/papers/genAC.pdf},
```
It is still not nice because a DOI and some kind of long term storage would be preferable. Perhaps Don would be willing to upload it to arxiv.org or zenodo.org. I've now emailed Don about this.

gap/Utils.gi
```diff
@@ -594,7 +594,85 @@ function(epsilon, d, q)
...
+    if d = 1 then
+
+        L1 := [[one]];
+        L2 := [[zeta]];
+        L3 := [[one]];
+
+    elif d = 2 and q = 3 then
+
+        L1 := [[-one, one], [-one, 0 * one]];
+        L2 := [[-one, one], [one, 0 * one]];
```
fingolfin, line 626: a matter of taste, but I usually just write it like this (seems more readable to me)
```suggestion
        L1 := one * [[-1, 1], [-1, 0]];
        L2 := one * [[-1, 1], [1, 0]];
```

gap/Utils.gi
```diff
@@ -594,7 +594,85 @@ function(epsilon, d, q)
...
+        L1 := NullMat(d, d, field);
+        L1[1, d] := one;
+        L1{[2..d]}{[1..d - 1]} := -IdentityMat(d - 1, field);
+
+        # In case q = 2, this matrix is just equal to L3, which
+        # makes sense because SL(d, 2) = GL(d, 2).
+        # In case q = 3, since L1 and L3 generate SL(d, 3)
+        # and [ GL(d, 3): SL(d, 3) ] = 2, adjoining any matrix 
+        # of determinant -1 to L1 and L2 will give GL(d, 3).
+        # Since we only want 2 generators, this matrix is
```
fingolfin, line 643:
```suggestion
        # Since we only want two generators, this matrix is
```

doc/ClassicalMaximals.bib
fingolfin: Don replied, he likes the idea of putting this on the arxiv, and even dug out the tex code for that. I expect it to appear there soon :-)

fingolfin (approved):

PR #104, opened 2022-01-21
Set quadratic forms

maxhauck:
Modify `ConjugateToSesquilinearForm` so that the quadratic form is also set at the end.

### General


fingolfin (approved):
Looks good to me, thanks

PR #109, opened 2022-01-29
Orthogonal C3

maxhauck:
Done!

### General


tst/standard/SemilinearMatrixGroups.tst
```diff
@@ -57,5 +57,21 @@ gap> TestMatricesInducingGaloisGroupOfGFQToSOverGFQ(2, 2);
 gap> TestMatricesInducingGaloisGroupOfGFQToSOverGFQ(5, 3);
 gap> TestMatricesInducingGaloisGroupOfGFQToSOverGFQ(4, 3);
 
+#
+gap> TestGammaLMeetOmega := function(epsilon, d, q, s)
+>   local G;
+>   G := GammaLMeetOmega(epsilon, d, q, s);
+>   Assert(0, IsSubsetOmega(epsilon, d, q, G));
+>   Assert(0, CheckSize(G));
```
fingolfin, line 65:
```suggestion
>   CheckIsSubsetOmega(epsilon, d, q, G);
>   CheckSize(G);
```
to match changes introduced in PR #108

fingolfin: You'll have to apply this change, else CI tests will fail

fingolfin (approved):

PR #110, opened 2022-02-08
Orthogonal C5

TristanPfersdorff:
All the stuff for the C5 subgroups in case O.

### General

### DIFF
--- a/gap/ClassicalMaximals.gi
+++ b/gap/ClassicalMaximals.gi
@@ -1123,10 +1123,15 @@ end);
 #       if numberOfConjugates = 8.
 BindGlobal("ConjugateSubgroupOmega",
 function(epsilon, n, q, G, numberOfConjugates)
-    local elementsToConjugate, result, powerSet, subset, exponent;
+    local elementsToConjugate, result, invariantFormRec,
+    powerSet, subset, exponent, conjugatedGroup;
 
-    if not numberOfConjugates in [2, 4, 8] then
-        ErrorNoReturn("<numberOfConjugates> must be one of 2, 4, 8");
+    if not numberOfConjugates in [1, 2, 4, 8] then
+        ErrorNoReturn("<numberOfConjugates> must be one of 1, 2, 4, 8");
+    fi;
+
+    if numberOfConjugates = 1 then
+        return [G];
     fi;
 
     elementsToConjugate := [SOMinusOmega(epsilon, n, q)];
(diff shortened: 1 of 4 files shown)


gap/ClassicalMaximals.gi
```diff
@@ -1138,19 +1143,79 @@ function(epsilon, n, q, G, numberOfConjugates)
...
+
+    listOfrs := List(PrimeDivisors(e));
+    result := [];
+
+    if epsilon = 0 then
+
+        # number of conjugates according to [KL90] Proposition 4.5.8 (I)
+        if 2 in listOfrs then
+            Append(result, ConjugateSubgroupOmega(epsilon, n, q, SubfieldOmega(0, n, p, e, QuoInt(e, 2), 0), 2));
+            Remove(listOfrs, 1);
```
fingolfin, line 1179: This kinda presupposes that 2 is always at position 1 in`listOfrs`. While this is true in practice here, it requires "non-local" knowledge of the code. So how about this instead?
```suggestion
            RemoveSet(listOfrs, 2);
```

gap/ClassicalMaximals.gi
```diff
@@ -1138,19 +1143,79 @@ function(epsilon, n, q, G, numberOfConjugates)
...
+                    numberOfConjugatesPlus := 1;
+                    numberOfConjugatesMinus := 1;
+                else
+                    if n mod 4 = 2 and p ^ QuoInt(e, 2) mod 4 = 1 then
+                        numberOfConjugatesPlus := 2;
+                        numberOfConjugatesMinus := 4;
+                    else
+                        numberOfConjugatesPlus := 4;
+                        numberOfConjugatesMinus := 2;
+                    fi;
```
fingolfin, line 1198: Personally I prefer to avoid nesting code more than necessary (a matter of taste, but I find it easier to grok the code this way)
```suggestion
                elif n mod 4 = 2 and p ^ QuoInt(e, 2) mod 4 = 1 then
                    numberOfConjugatesPlus := 2;
                    numberOfConjugatesMinus := 4;
                else
                    numberOfConjugatesPlus := 4;
                    numberOfConjugatesMinus := 2;
                fi;
```

fingolfin (approved):
Looks good to me, I have some nitpicky style comments which you can ignore or address as you like

TristanPfersdorff:
Nitpicks are implemented. I also noticed that my tests were apparently so terrible that a large chunk of code was not covered, but that is all fixed now to. This is ready for merging.

gap/SubfieldMatrixGroups.gi
```diff
@@ -334,3 +334,107 @@ function (d, p, e, f)
...
+    zeta := PrimitiveElement(field);
+    lambda := zeta ^ QuoInt(q0 + 1, 2);
+    lambdaInv := lambda ^ -1;
+    m := QuoInt(d, 2);
+
+    if epsilon_0 = 1 then
+
+        A := lambda * IdentityMat(d, field);
+        A{[m + 1..d]}{[m + 1..d]} := lambdaInv * IdentityMat(m, field);
+        F := StandardOrthogonalForm(epsilon, d, q0).F;
```
fingolfin, line 416: This block apparently is not covered (according to Codecov) ?

fingolfin (approved):

gap/SubfieldMatrixGroups.gi
TristanPfersdorff: Ahh, I spoke too soon, it's fixed now (for real this time).

PR #115, opened 2022-03-30
Orthogonal Maximals

TristanPfersdorff:
Tests for the orthogonal functions in `gap/ClassicalMaximals.gi`. The tests for `n = 9` will work once the C7-PR is merged (at least that's the idea). Also, the C6-stuff never gets tested since class C6 is empty for `n <= 12` in case O, which is a bit sad.


fingolfin:
Tests still fail

fingolfin (approved):

PR #116, opened 2022-03-31
Add AntidiagonalHalfOneMat helper

fingolfin:


TristanPfersdorff (approved):
Long overdue, this is much more readable! Interestingly we seemingly had the 1s and -1s flipped in `gap/ClassicalNormalizer.gi`, which didn't cause any issues because we were just off by a root of unity.

PR #128, opened 2024-07-07
Fix size issue in `SpStabilizerOfIsotropicSubspace`

### LINKED ISSUE #127: bug in ClassicalMaximalsGeneric("S",4,5);
It looks as if size are set at a certain point in the creation of these groups, and the example below shows at least one error in this. 
```
gap> groups := ClassicalMaximalsGeneric("S",4,5);
[ <matrix group of size 60000 with 6 generators>, 
  <matrix group of size 60000 with 4 generators>, 
  <matrix group of size 28800 with 3 generators>, 
  <matrix group of size 960 with 3 generators>, 
  <matrix group of size 31200 with 3 generators>, 
  <matrix group of size 1440 with 3 generators>, 
  <matrix group of size 1920 with 8 generators> ]
gap> grp := groups[2];
<matrix group of size 60000 with 4 generators>
gap> Size(Group(GeneratorsOfGroup(grp)));
300000
```

TristanPfersdorff:
Pretty much everything I mentioned in [here](https://github.com/gap-packages/ClassicalMaximals/issues/127#issuecomment-2037870811), tackling #127.


fingolfin (approved):
This looks but the base CI is badly broken. Ideally this is fixed first and then we merge this PR when its CI passes... I'll see what I can do

PR #145, opened 2026-02-17
Construct Class C9 Maximal Subgroups up to Dimension 12

chseeger:
This pull request contains the implementation of Class $\mathcal{C}_9$ maximal subgroups of classical groups up to and including dimension 12.

---
General remarks:
- Subgroup sizes are only set and tested for the "default input parameters", i.e. when computing subgroups of the relevant quasisimple group with `special = false`, `general = false` and `normaliser = false`.
- The implementation relies on `recog 1.4.4dev`.
- Documentation is incomplete and will require a dedicated effort ...


fingolfin:
Unfortunately GitHub struggles with displaying the content of this PR which impedes review.

I have therefore cherry-picked your commit ca6b1b714c92aae58a9d4ec080bdeed95bfedc5e which adds the `data` dir onto the `main` branch.

Please rebase this PR now on latest master and force-push (I can also do that but I don't want to risk messing up your work flow).

Then afterwards I should be able to properly review things and leave comments if necessary.

gap/AlmostSimpleCrossCharacteristic.gi
```diff
@@ -0,0 +1,371 @@
...
+            od;
+            for j in sgens do
+                # Attempt to append this automorphism to AC[i]
+                a := AI[j];
+                phi := GroupHomomorphismByImages(F, AC[i], [AC[i].1, AC[i].2]);
+                M2 := GModuleByMats([phi(a[1]), phi(a[2])], GF(q));
+                iso := MTX.IsomorphismModules(C[i], M2);
+                Assert(0, iso <> fail);
+                # adjust by scalar to make iso fix form.
+                scal := (iso * form * TransposedMat(EntrywisePowerMat(iso, rq))* form^-1)[1,1];
```
fingolfin:
```suggestion
                scal := (iso * form * HermitianConjugate(iso, rq)* form^-1)[1,1];
```

gap/AlmostSimpleDefiningCharacteristic.gi
```diff
@@ -0,0 +1,1182 @@
...
+        g := PseudoRandom(G);
+        f := PrimePowersInt(Order(g));
+        f := List([1..Length(f)/2], i -> [f[2*i-1], f[2*i]]);
+    until Size(f) > 1 or (Size(f) = 1 and f[1,1] <> 2);
+    if f[1,1] = 2 then
+        g := g^(2^f[1,2]);
+    fi;
+    H := GroupByGenerators([g, g^PseudoRandom(G)]);
+    while not IsAbsolutelyIrreducible(H) do
+        H := GroupByGenerators(Concatenation(GeneratorsOfGroup(H), [g^PseudoRandom(G)]));
```
fingolfin, line 1147:
```suggestion
        H := ClosureGroup(H, g^PseudoRandom(G));
```

gap/AlmostSimpleDefiningCharacteristic.gi
```diff
@@ -0,0 +1,1182 @@
...
+    od;
+    while false do  # TODO while not RecogniseClassical(H)
```
fingolfin: Maybe this is meant?
```suggestion
    while RecogniseClassical(H).isOmegaContained <> true then
```
However, in older recog versions, the member actually is called `isSOContained` (IIRC)... hrmmm

gap/AlmostSimpleDefiningCharacteristic.gi
```diff
@@ -0,0 +1,1182 @@
...
+    G := GroupByGenerators(Concatenation(Hgi, [x]));
+    # Conjugate G s.t. it preserves our standard form *up to a scalar*.
+    # Ideally, the function ConjugateToStandardForm should be extended to include this
+    # functionality.
+    oldForm := PreservedSesquilinearForms(G)[1];
+    newForm := BilinearFormByMatrix(InvariantBilinearForm(Omega(1, 8, q^2)).matrix, GF(q^2));
+    baseChangeMatrix := BaseChangeToCanonical(oldForm)^-1 * BaseChangeToCanonical(newForm);
+    G := MatrixGroup(GF(q^2), List(GeneratorsOfGroup(G), g -> g ^ baseChangeMatrix));
+    return GroupByGenerators(Concatenation(GeneratorsOfGroup(G),
+                                           [PrimitiveElement(GF(q^2)) * IdentityMat(8, GF(q^2))]));
```
fingolfin:
```suggestion
    G := G ^ baseChangeMatrix;
    return ClosureGroup(G, PrimitiveElement(GF(q^2)) * IdentityMat(8, GF(q^2)));
```
Do you really just need a `PrimitiveElement` or should this perhaps use `PrimitiveRoot` resp. `Z(q^2)`?

gap/AlmostSimpleDefiningCharacteristic.gi
```diff
@@ -0,0 +1,1182 @@
...
+    x := MTX.IsomorphismModules(M, Mi);
+    if x = fail then
+        # "try other one"
+        Hgc2 := List(Hg, h -> h^(ngo * go));
+        Hgc2i := List(Hgc2, h -> phi(h));
+        M2i := GModuleByMats(Hgc2i, GF(q^2));
+        x := MTX.IsomorphismModules(M, M2i);
+        Assert(0, x <> fail);
+    fi;
+    G := GroupByGenerators(Concatenation(Hgi, [x]));
```
fingolfin, line 1171:
```suggestion
    G := ClosureGroup(Hgi, x);
```

fingolfin (approved):

gap/AlmostSimpleDefiningCharacteristic.gi
chseeger: > Do you really just need a `PrimitiveElement` or should this perhaps use `PrimitiveRoot` resp. `Z(q^2)`?

Agreed, `PrimitiveRoot` should be used here, not `PrimitiveElement`. I suspect the same issue appears elsewhere in the package; I've replaced it where necessary in this branch.

gap/AlmostSimpleDefiningCharacteristic.gi
chseeger: Thanks, I wasn't aware of `ClosureGroup`.
This probably isn’t the best place to mention it, but `ClosureGroup` can be significantly slower than the explicit `GroupByGenerators(Concatenation(...))` approach. It triggers a `NiceMonomorphism` call that is quite expensive for certain groups:
```gap
gap> ClosureGroup(Sp(4,23), NormSpMinusSp(4,23));; time;
24207
```
Therefore I would wait with a global replacement (across the entire package) and perhaps do that in a separate PR.

PR #151, opened 2026-03-30
Simplify form detection code

fingolfin:


chseeger:
If I remember correctly, the reason I did not make this simplifying change some time ago was the behavior of `MTX.InvariantQuadraticForm`. Before this change, `CM_QuadraticForm` returned an upper triangular matrix[^1], following the convention used in the `Forms` package and in Magma (and parts of the implementation of Class $\mathcal{C}_9$ relied on this, at least in characteristic 2 ...).

For example in characteristic 2, `CM_QuadraticForm` now returns a lower triangular matrix instead[^1].

Of course, this change in behavior is perfectly reasonable, since we neither had a formal specification of `CM_QuadraticForm` in the documentation nor explicit tests enforcing the previous convention. However, we should document the new behavior and explicitly point out the difference compared to the `Forms` package.

This is also relevant because, for certain maximal subgroups, we explicitly set invariant forms using those stored in the GAP groups `Omega(epsilon, n, q)`, etc.; see `ConjugateToStandardForm`:

```gap
gap> G := Omega(1,4,2);;
gap> M := NaturalGModule(G,GF(2));;
gap> Display(InvariantQuadraticForm(G).matrix);
 . . . 1
 . . 1 .
 . . . .
 . . . .
gap> Display(CM_QuadraticForm(G,GF(2)));
 . . . 1
 . . 1 .
 . . . .
 . . . .
gap> Display(MTX.InvariantQuadraticForm(M));
 . . . .
 . . . .
 . 1 . .
 1 . . .
gap>
```
Ultimately, however, `CM_QuadraticForm` and `CM_BilinearForm` are just internal helper functions that are not really intended for the user.

[^1]: unless a matrix has already been set in the record `InvariantQuadraticForm` that does not have this property ...

PR #154, opened 2026-05-02
Fix and generalize geometric maximal subgroups of orthogonal groups

chseeger:
This PR resolves [#150](https://github.com/gap-packages/ClassicalMaximals/issues/150) and addresses some issues in the construction of geometric maximal subgroups of orthogonal groups. For details, see [#150](https://github.com/gap-packages/ClassicalMaximals/issues/150).
After this PR, the situation _for orthogonal groups_ is as follows (I will update the relevant section in the README once the documentation has been expanded):
- The returned lists should be correct and complete for all dimensions $5\leq n \leq 12$.
- For $n=3,4$, missing subgroups in $\mathcal{C}_1$ are explicitly flagged via the new info class `InfoClassicalMaximals`.
- Classes $\mathcal{C}_2$ and $\mathcal{C}_4$ should be complete for all dimensions $n \geq 3$.
- For $3 \leq n \leq 12$ and prime powers $q < 20$, the tests confirm that the number of maximal subgroups agrees with Magma (except for the known missing $\mathcal{C}_1$ subgroups for $n=3,4$, and a duplicate in Magma's C4 output).

### DIFF
--- a/tst/standard/ClassicalMaximals.tst
+++ b/tst/standard/ClassicalMaximals.tst
@@ -13,24 +13,20 @@ gap> MaximalSubgroupClassRepsSpecialLinearGroup(2, 2);
 Error, SL(2, 2) and SL(2, 3) are soluble
 gap> MaximalSubgroupClassRepsSpecialUnitaryGroup(3, 5, [0]);
 Error, <classes> must be a subset of [1..9]
-#@if IsBound(CLASSICAL_MAXIMALS_RUN_BROKEN_TESTS)
-gap> MaximalSubgroupClassRepsSpecialUnitaryGroup(2, 3);  # FIXME: Unstable output formatting (line breaks) in error message
-Error, We assume <n> to be greater or equal to 3 in case 'U' since SU(2, q) and SL(2, q) are isomorphic
-#@fi
+gap> MaximalSubgroupClassRepsSpecialUnitaryGroup(2, 3);
+Error, <n> must be at least 3 in case 'U' since SU(2, q) and SL(2, q) are isom\
+orphic
 gap> MaximalSubgroupClassRepsSpecialUnitaryGroup(3, 2);
 Error, PSU(3, 2) is soluble
 gap> MaximalSubgroupClassRepsSymplecticGroup(4, 3, [0]);
 Error, <classes> must be a subset of [1..9]
-#@if IsBound(CLASSICAL_MAXIMALS_RUN_BROKEN_TESTS)
-gap> MaximalSubgroupClassRepsSymplecticGroup(2, 3);  # FIXME: Unstable output formatting (line breaks) in error message
-Error, We assume <n> to be greater or equal to 4 in case 'S' since Sp(2, q) and SL(2, q) are isomorphic
-#@fi
+gap> MaximalSubgroupClassRepsSymplecticGroup(2, 3);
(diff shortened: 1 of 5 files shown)


fingolfin:
Thanks a lot!!!

Question: If this PR resolves issue #150 (and hence closes it), we don't have any issue left pointing out the problem with $\mathcal{C}_1$. Wouldn't it be more appropriate that this PR resolves most of that issue? Anyway, I don't mind closing #150 as long as there is some record about what is still broken / missing.

tst/standard/ClassicalMaximals.tst
```diff
@@ -1,4 +1,46 @@
...
+gap> MaximalSubgroupClassRepsSpecialLinearGroup(2, 4, [0]);
+Error, <classes> must be a subset of [1..9]
+gap> MaximalSubgroupClassRepsSpecialLinearGroup(2, 2);
+Error, SL(2, 2) and SL(2, 3) are soluble
+gap> MaximalSubgroupClassRepsSpecialUnitaryGroup(3, 5, [0]);
+Error, <classes> must be a subset of [1..9]
+#@if IsBound(CLASSICAL_MAXIMALS_RUN_BROKEN_TESTS)
+gap> MaximalSubgroupClassRepsSpecialUnitaryGroup(2, 3);  # FIXME: Unstable output formatting (line breaks) in error message
+Error, We assume <n> to be greater or equal to 3 in case 'U' since SU(2, q) and SL(2, q) are isomorphic
+#@fi
```
fingolfin: This sounds like it suffers from that fact that the error message will be wrapped by GAP based on the terminal width.

But the wrapping should actually be deterministic when `Test` runs. I think we set it to 80 columns or so.

So something like this should work (though the wrapping may have to be adjusted slightly)
```suggestion
gap> MaximalSubgroupClassRepsSpecialUnitaryGroup(2, 3);  # FIXME: Unstable output formatting (line breaks) in error message
Error, We assume <n> to be greater or equal to 3 in case 'U' since SU(2, q) an\
d SL(2, q) are isomorphic
```

fingolfin: Also ideally the message would start in lower case ("Error, We" -> "Error, we").

It could also be shorter:
> Error, <n> must be at least 3 in case 'U' as SU(2,q) and SL(2,q) are isomorphic

fingolfin (approved):
Looks good to me, some very minor quibbles

chseeger:
> Question: If this PR resolves issue #150 (and hence closes it), we don't have any issue left pointing out the problem with $\mathcal{C}_1$. Wouldn't it be more appropriate that this PR resolves most of that issue? Anyway, I don't mind closing #150 as long as there is some record about what is still broken / missing.

Thank you for reviewing and pointing this out. I've opened a dedicated issue ([#155](https://github.com/gap-packages/ClassicalMaximals/issues/155)) that explicitly documents the $\mathcal{C}_1$ problem for $n=3,4$.

Given that [#150](https://github.com/gap-packages/ClassicalMaximals/issues/150) bundles several independent issues (essentially all addressed by this PR) and does not clearly highlight this particular problem, I suggest closing it once this PR is merged.

I'll also update the README in a follow-up (the TODO section is quite outdated) to document such limitations there. In addition, the upcoming documentation PR (to be submitted in the next few days) will briefly mention this case as well.

fingolfin:
Wonderful, thanks a lot

PR #162, opened 2026-05-07
Re-enable many formerly broken tests

fingolfin:
Many issues were fixed in recog. Unfortunately one new one cropped up.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -35,7 +35,7 @@ jobs:
           gap-version: ${{ matrix.gap-version }}
       - uses: gap-actions/install-pkg@v1
         with:
-          packages: gap-packages/recog@devel gap-packages/orb@devel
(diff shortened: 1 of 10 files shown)


fingolfin:
Until recently, GAP did not clear the options stack after an error. In `master` and the upcoming GAP 4.16.0, we are not doing it thanks to https://github.com/gap-system/gap/pull/6355

So if there is an unexpected error, we may end up with multiple values on the stack. Alas, where is that happening?

Looking at the code now, I also see that perhaps there is a misunderstanding going on as to how options work:

GAP global options are automatically and recursively passed on to *every* function called by a function which options set on. So invocations like this:
```
            S := ModularReductionOfIntegralLattice(LR, q : normaliser := normaliser);
```
are really kinda unnecessary, you could just as well do
```
            S := ModularReductionOfIntegralLattice(LR, q);
```
Passing the `normaliser` again just means that it ends up multiple times on the options stack, as you observed. Which should be harmless, just redundant.

PR #165, opened 2026-05-16
Handle low-dimensional cases in orthogonal C1 subgroups construction

### LINKED ISSUE #155: Handle C1 (maximal parabolic) subgroups in `Omega(n,q)`for `n < 5`
Since [#154](https://github.com/gap-packages/ClassicalMaximals/pull/154$0), we allow orthogonal groups in dimensions $n \geq 3$ (previously $n \geq 7$), using the tables in [BHR13] to filter out non-maximal subgroups. The constructors themselves were left unchanged.

The only remaining gap is in class $\mathcal{C}_1$: The constructor `OmegaStabilizerOfIsotropicSubspace` implements Lemma 4.2 in [HR10], but this construction appears to fail for $n < 5$.

This is not surprising: [HR10] assumes $n \geq 7$, noting that for $n < 7$, $\Omega^\varepsilon(n,q)$ is either not quasisimple or isomorphic to a classical group of different type. (This was also the reason for the previous lower bound $n \geq 7$.)

In practice, however, we now support these dimensions (consistent with [BHR13] and Magma), so the missing constructions become visible.
The issue affects quasisimple groups `Omega(3,q)` and `Omega(-1,4,q)`. At the moment, we show a warning (via `InfoClassicalMaximals`) indicating that the $\mathcal{C}_1$ subgroup of type $P_k$ is missing, and we skip constructing it.

Possible approaches include:
- implementing ad hoc constructions for $n = 3, 4$,
- or handle these cases via identifications with other classical groups.

chseeger:
Reorganises the p-core construction in `OmegaStabilizerOfIsotropicSubspace` to correctly handle the low-dimensional orthogonal cases $(\varepsilon,n,k)=(0,3,1)$ and $(-1,4,1)$.

Closes [#155](https://github.com/gap-packages/ClassicalMaximals/issues/155).

### DIFF
--- a/gap/ReducibleMatrixGroups.gi
+++ b/gap/ReducibleMatrixGroups.gi
@@ -652,7 +652,7 @@ function(epsilon, d, q, k)
         if IsEvenInt(q) then
             gamma := FindGamma(q);
         else
-            xi := PrimitiveElement(GF(q ^ 2));
+            xi := PrimitiveRoot(GF(q ^ 2));
             gamma := xi ^ (q + 1) * (xi + xi ^ q) ^ -2;
         fi;
         eta := (1 - 4 * gamma) ^ -1;
--- a/gap/ReducibleMatrixGroups.gi
+++ b/gap/ReducibleMatrixGroups.gi
@@ -546,9 +546,8 @@ function(epsilon, d, q, k)
         ErrorNoReturn("<k> must not be equal to <m> for <epsilon> = -1");
     fi;
 
-    # the construction referenced above fails for d < 5
-    if d < 5 then
-        ErrorNoReturn("<d> must be at least 5");
+    if d < 3 then
+        ErrorNoReturn("<d> must be at least 3");
     fi;
 
     field := GF(q);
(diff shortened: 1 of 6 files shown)


gap/ReducibleMatrixGroups.gi
```diff
@@ -625,41 +627,36 @@ function(epsilon, d, q, k)
...
+    elif d = 2*k + 1 then  # odd case
+
+        Add(gens, IdentityMat(d, field) + MatrixByEntries(field, d, d, [[k + 1, 1, one], [d, d - k, -one], [d, 1, -one / 2]]));
+
+    elif d = 2*k + 2 and epsilon = -1 then
+
+        if IsEvenInt(q) then
+            gamma := FindGamma(q);
+        else
+            xi := PrimitiveElement(GF(q ^ 2));
```
fingolfin: Do you need a primitive element, or a primitive root (which is a stronger requirement)

chseeger: Agreed, `PrimitiveRoot` should be used here. I would't be surprised if the same mistake shows up elsewhere across the package, so I'll check for that later (but not in this PR).

fingolfin: Thanks! See also https://github.com/gap-system/gap/issues/5967

fingolfin (approved):

PR #167, opened 2026-05-16
Correct classification of sesquilinear forms

fingolfin:
Also clarify a comment about historic chouces.

### DIFF
--- a/doc/theory.autodoc
+++ b/doc/theory.autodoc
@@ -65,23 +65,17 @@ $g$ is a **similarity**.
 
 @Subsection Classification of sesquilinear forms
 @SubsectionLabel Theory:ClassicalGroups:FormsClassification
-By the following fundamental result (<Cite Key="BHR13" Where="Theorem 1.5.13" />)
-the $\sigma$-sesquilinear forms fall into a small number of types
-(up to similarity):
-
-Let $\beta$ be a $\sigma$-sesquilinear form on $V$, and let
-$\lambda \in K^\times$, $\tau \in {\rm Aut}(K)$ satisfy
-$\lambda \beta(u, v)^\tau = \beta(u, v)$ for all $u, v$. Then, up to
-similarity, one of the following holds.
+Let $\beta$ be a $\sigma$-sesquilinear form on $V$. Then following
+<Cite Key="BHR13" Where="Theorem 1.5.13" />, we focus on the following four classes
+of such forms:
 <Enum>
  <Item>$\beta = 0$.</Item>
- <Item>$\sigma = 1$, $\lambda = -1$ and $\beta(v, v) = 0$ for all $v$.
-       $\beta$ is called **alternating** or **symplectic**.</Item>
- <Item>$\sigma^2 = 1 \neq \sigma$, $\lambda = 1$
-       (i.e. $\beta(v, u) = \beta(u, v)^\sigma$).
-       $\beta$ is called **$\sigma$-Hermitian** or **unitary**.</Item>
- <Item>$\sigma = 1$, $\lambda = 1$ (i.e. $\beta(v, u) = \beta(u, v)$).
-       $\beta$ is **symmetric bilinear**.</Item>
+ <Item>$\sigma = 1$, $\beta(u, v) = -\beta(v, u)$ for all $u, v$, and $\beta(v, v) = 0$ for all $v$.
+       Then $\beta$ is called **alternating** or **symplectic**.</Item>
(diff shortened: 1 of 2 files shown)


doc/theory.autodoc
```diff
@@ -69,19 +69,19 @@ By the following fundamental result (<Cite Key="BHR13" Where="Theorem 1.5.13" />
 the $\sigma$-sesquilinear forms fall into a small number of types
 (up to similarity):
 
-Let $\beta$ be a $\sigma$-sesquilinear form on $V$, and let
-$\lambda \in K^\times$, $\tau \in {\rm Aut}(K)$ satisfy
-$\lambda \beta(u, v)^\tau = \beta(u, v)$ for all $u, v$. Then, up to
+Let $\beta$ be a $\sigma$-sesquilinear form on $V$, and suppose there is
+$\lambda \in K^\times$ such that
+$\lambda \beta(u, v)^\sigma = \beta(u, v)$ for all $u, v$. Then, up to
```
chseeger: Thank you for pointing this out. The current version in the manual is definitely incorrect.

In [BHR13, Theorem 1.5.13] (a slight variation of the Birkhoff-von Neumann theorem), the premise of this classification theorem is that " $\beta$ is a quasi-symmetric $\sigma$-sesquilinear form".

My original intention behind introducing $\tau$ was to state the classification theorem correctly without having to formally define "quasi-symmetric" (since its definition is not uniform in the literature and the term is only used in this specific place).

For context, the relevant definitions from [BHR13, Definition 1.5.8] are: A $\sigma$-sesquilinear form $\beta$ is
- quasi-symmetric if there exist $\lambda \in F^\times$ and $\tau \in {\rm Aut}(F)$ such that $\beta(v, u) = \lambda \beta(u, v)^\tau$ for all $u, v \in V$.
- reflexive if it satisfies the property that $\beta(u, v) = 0$ if and only if $\beta(v, u) = 0$.

The current version in the manual is clearly wrong because I forgot the argument swap $(u,v) \mapsto (v,u)$ in $\beta(v, u) = \lambda \beta(u, v)^\tau$. I think if we correct this, the statement should be accurate again?

Consequently, I think the proposed change is also not entirely correct, as it misses the argument swap as well (for instance, in the symplectic case outside characteristic 2, this would imply that $\beta = 0$).

As a side note, the more well-known version of the Birkhoff-von Neumann theorem requires the form to be "reflexive non-degenerate" instead of "quasi-symmetric". As an alternative, we could introduce the term "reflexive" and present that version instead ...

fingolfin: Indeed, I noticed the missing swap but then ultimately forgot to fix it *sigh*.

Anyway, I've rewritten it again to not claim a classification (though it is mentioned in the title), but just list the form types. That is simpler and avoids introducing notions like quasi-symmetric.

I've also re-arranged sections a bit

doc/theory.autodoc
```diff
@@ -65,44 +60,46 @@ $g$ is a **similarity**.
...
- <Item>$\sigma = 1$, $\lambda = -1$ and $\beta(v, v) = 0$ for all $v$.
-       $\beta$ is called **alternating** or **symplectic**.</Item>
- <Item>$\sigma^2 = 1 \neq \sigma$, $\lambda = 1$
-       (i.e. $\beta(v, u) = \beta(u, v)^\sigma$).
-       $\beta$ is called **$\sigma$-Hermitian** or **unitary**.</Item>
- <Item>$\sigma = 1$, $\lambda = 1$ (i.e. $\beta(v, u) = \beta(u, v)$).
-       $\beta$ is **symmetric bilinear**.</Item>
+ <Item>$\sigma = 1$, $\beta(u, v) = -\beta(v, u)$ for all $u, v$, and $\beta(v, v) = 0$ for all $v$.
+       Then $\beta$ is called **alternating** or **symplectic**.</Item>
+ <Item>$\sigma^2 = 1 \neq \sigma$ and $\beta(v, u) = \beta(u, v)^\sigma$) for all $u, v$.
```
chseeger:
```suggestion
 <Item>$\sigma^2 = 1 \neq \sigma$ and $\beta(v, u) = \beta(u, v)^\sigma$ for all $u, v$.
```

chseeger (approved):
Thank you! Looks very good from my side.
(with a nitpick suggested change for a closing parenthesis)

doc/theory.autodoc
chseeger: Yes, I agree that avoiding a rigorous classification theorem and just listing the form types is much simpler and entirely sufficient.

PR #176, opened 2026-07-23
Generalize the geometric subgroup constructions to arbitrary dimensions and implement the `all` option

chseeger:
- Generalize the top-level implementations of the C1-C8 constructions to arbitrary dimensions by following the tables in [KL90] (previously, a few case distinctions were hard-coded up to dimension 12). The subgroup constructors themselves are unchanged.
- Implement the `all` option for the C1-C8 constructions (the C9 construction already supports this option).

The `all` option remains undocumented for now.

A few tests are temporarily disabled because of open issues in `recog` (including issue [#510](https://github.com/gap-packages/recog/issues/510)); otherwise, I suspect the Codecov workflow would not produce a coverage report.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,15 +11,13 @@ Translation of magma `ClassicalMaximals` to GAP. For resources see
 ### Implementation status
 
 #### Geometric maximal subgroups (Aschbacher Classes C1-C8)
-- **Type L**: Complete for dimensions 2-12
-- **Type U**: Complete for dimensions 3-12
-- **Type S**: Complete for dimensions 4-12
-- **Type O**: Complete for dimensions 3-12
-    - **C2 & C4**: Complete for all dimensions
+- Complete for types L, U, S, O in all dimensions
+- Supported options (via option records, undocumented):
+    - `all` (see below)
 
 #### Almost simple groups (Class S)
 - Complete for types L, U, S, O in dimensions up to 12
-- **Supported options** (via option records, undocumented):
+- Supported options (via option records, undocumented):
     - `all`: Conjugacy classes under the full automorphism group of the simple classical group
     - `novelties`: Intersections of novelty maximal subgroups with the quasisimple group
     - `special`: Normalisers in SO(n,q)
@@ -32,14 +30,13 @@ Translation of magma `ClassicalMaximals` to GAP. For resources see
 #### Testing & verification
 - Verification of stored bilinear/sesquilinear/quadratic forms
 - Group size checks via the `recog` package
-- Cross-checks against tables in [BHR13] for the number of maximal subgroups
-    - Class S & orthogonal geometric subgroups: Tests also against Magma's `ClassicalMaximals`
+- Cross-checks against tables in [BHR13] and against Magma's `ClassicalMaximals`
+  for the number of maximal subgroups
 
 ### Roadmap / TODO
 
 #### Geometric maximal subgroups (Aschbacher Classes C1-C8)
-- Generalize other Aschbacher classes to work for all dimensions
-- Implement `all`, `novelties`, `special`, `general`, `normaliser` for all geometric classes
+- Implement `novelties`, `special`, `general`, `normaliser` for all geometric classes
 
 #### Almost simple groups (Class S)
 - Extend implementation beyond dimension 12 (for comparison: Magma covers dimensions up to 17)
--- a/doc/intro.autodoc
+++ b/doc/intro.autodoc
@@ -25,7 +25,7 @@ in principle, sufficient information to determine these subgroups up to dimensio
 However, explicit computations have so far only been carried out up to dimension 17,
 <Cite Key="BHR13" /> does so up to dimension 12 and for these dimensions our returned
 lists should therefore be complete.
-For higher dimensions we give no completeness guarantee; in particular,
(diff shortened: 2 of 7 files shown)


fingolfin:
https://github.com/gap-packages/recog/issues/510 is now fixed, perhaps you could try re-enabling the tests?

chseeger:
Yes, the recent fixes in `recog` allow us to run these tests again, so I have re-enabled them.

fingolfin (approved):

fingolfin:
Nice! Folding the hard-coded per-dimension cases into the generic functions makes this much easier to follow.

A few things Claude AI noticed (lightly edited by me):

### Blocking

**1. API break in the per-type functions.** The list form still works for `ClassicalMaximalsGeneric`, but not for the four public `MaximalSubgroupClassReps*` functions:

```
gap> MaximalSubgroupClassRepsSpecialLinearGroup(4, 5, [1,2]);
Error, Record IsBound: <rec> must be a record (not a plain list of cyclotomics)
```

These are declared in `ClassicalMaximals.gd`, so this breaks existing callers. Please apply the same normalisation there, or at least raise a clear error.

**2. The caller's record is mutated.** `opts.classes := [1..9]` writes into the record passed in, which can cause all kinds of troubles. Just do `opts := ShallowCopy(opts)`.

**3. Class validation silently dropped.** `MaximalSubgroupClassRepsSymplecticGroup(4, 3, [0])` used to error; now invalid classes pass unchecked (`rec(classes := [0])` just returns `[]`). The removed test lines at the top of `ClassicalMaximals.tst` acknowledge this. But why? Validation can catch nasty errors and costs us virtually nothing, does it?

### Worth a second pair of eyes on [KL90]

These change results and are guarded only by the new `Length(...)` values:

- C7 orthogonal, symplectic-tensor-induced branch: `numberOfConjugates := gcd` → `2 * gcd`; `q mod 4 = 3` → `q mod 4 = 1` (Prop. 4.7.7); `primeDivs[1] = 3` → `primeDivs[1] = 2` (Prop. 4.7.5 exclusion). The `m mod 4 = 4` → `m mod 4 = 0` fix is clearly right; the others I could not check.
- C6 orthogonal: the old exclusion `q ≡ 3, 5 (mod 8)` applied to every `n`, the new one only to `n = 8`. So for `n = 16, 32, …` with `q ≡ ±3 (mod 8)` the subgroup is now included where it previously was not. Correct if the exclusion really stems only from the `n = 8` triality novelty — worth confirming against Table 3.5.E.

### Minor

- `MaximalSubgroupClassRepsSpecialUnitaryGroup` still computes `generatorGUMinusSU := GUMinusSU(n, q)` and declares `subfieldGroup`, `numberOfConjugates`; all three are now unused.
- `doc/intro.autodoc` says "only maximal subgroups of geometric type are considered", while `ClassicalMaximals.gd` says "complete except for class C9". The `.gd` wording is the more accurate one; I'd use it in both places.

chseeger:
Thank you for having a look at this. I went through the points and made the following changes.

### Blocking
1. **API break in the per-type functions.** The normalisation from lists to records for `classes` in `ClassicalMaximalsGeneric` was the main reason why I did not merge this without another look. The motivation behind my change was that the four `MaximalSubgroupClassReps*` functions are not documented, and that supporting both list and record input makes these functions less simple. But I agree that breaking the existing list input is maybe not what we want. As suggested, I've applied the same normalisation there as in `ClassicalMaximalsGeneric`.
2. **The caller's record is mutated.** Thanks, fixed by using `ShallowCopy(opts)` before modifying it.
3. **Class validation silently dropped.** The class validation got lost when making the normalisation changes (see 1.). I've re-added the validation and the corresponding tests.

### Worth a second pair of eyes on [KL90]
- C7 orthogonal, symplectic-tensor-induced branch: (Note that the previous code was not covered by any tests at all)
  - `numberOfConjugates := gcd` → `2 * gcd`: double-checked against Proposition 4.7.5 (I)
  - `q mod 4 = 3` → `q mod 4 = 1` (Prop. 4.7.7): double-checked against Proposition 4.7.7 (I)
  - `primeDivs[1] = 3` → `primeDivs[1] = 2` (Prop. 4.7.5 exclusion): double-checked against Proposition 4.7.5 and Table 3.5.E, Column IV

- C6 orthogonal: For `n = 8`, `p = q ≡ ±3 (mod 8)`, the C6 subgroup of type `2_+^{1+2m}.O_{2m}^+(2)` is not maximal in `Ω_8^+(q)`, but a novelty, which we do not include by default. This is the triality case described in [Kl87, Proposition 2.4.10] and [BHR13, Table 8.50, Novelty N5]. Looking at this again, I am not completely sure whether further novelties arise for `n = 32, 64, …`. According to Rogers' PhD thesis, this is not the case for `n = 16`, and the implementation in Magma also treats only `n = 8` as a special case.

### Minor
- Removed the unused variables in `MaximalSubgroupClassRepsSpecialUnitaryGroup`.
- Changed the wording in `doc/intro.autodoc` to match `ClassicalMaximals.gd`.

[Kl87] P. B. Kleidman. The maximal subgroups of the finite 8-dimensional orthogonal groups $P\Omega^+\_8(q)$ and of their automorphism groups. *J. Algebra* **110** (1987), 173–242.

